PR #11, opened 2022-12-09
Back/initial db merge

Riul1999:
# Extended PR

This PR extended the Initial DB pull request below. The main goal was to enable database and dataloaders.

This PR purged most of old database and endpoint work, because it needs to be reconsidered. Only most basic endpoints such as dataset upload and dataset database are online.

**This PR will break most of frontend usability and endpoints. It is really important to reconsider how the calls are made, which endpoints are necessary and what parameters are given to the backend**

## Changes

- Database name is now DashAI.sqlite from dashAI.sqlite
- Merged with last PR "Change execution working directory, meet pep8 and update project structure"
- Renamed session back to db because of the call `db.session.query...` was `session.session.query...`
- Removed previous example_datasets that wont work with the new dataloaders. Next dataloaders PR will contain new CSV and JSON example datasets.
- main.py was purged of most endpoints.
- Frontend shouldn't need an endpoint to check if api is online. Removed /info.

### `back/routers/`
- Created routers to decrease main.py complexity
- Now dataset endpoint follow RESTful structure
- experiment route is a placeholder. Expermients need to be reconsidered with an orchestrator.

### `back/dataloaders/classes`
- Implemented multiple dataloaders from the abstract class DataLoader, where each dataloader should implement load\_data method
- DataLoader is responsible for split\_dataset and set\_classes

### `back/dataloaders/dataloader_schemas` & `back/dataloaders/params_schema`
- These jsons contain information that is shown on the frontend. Maybe on next iteration switch them to python dictionaries
- Translation is outdated and on newest branches of dataloaders its done.

Old changes are below

---
# Initial DB
Initial usage of Sqlite DB, now the experiment, task and executions are stored in a sqlite file to store the progress of an experiment. The results are also stored in the execution entity.

### `back/Database/models.py`
- Created classes Experiment and Execution to store the experiment and the executions in the DB.
- Method get_results from Experiment class returns all the results for every execution associated to the Experiment.
- Created class Dataset to store dataset instead of requesting the user every time he wants to make an experiment.

### `back/models/classes`
- Rename SklearnLikeModel
- Update method save in SklearnLikeModel to store files instead of return the bytes.

### `back/TaskLib/task/taskMain.py`
- Add methods save and load to store and load tasks from the DB.
- Now set_executions returns a single execution instead of storing all the executions.
- Now run_experiments returns the results dictionary instead of storing it inside the task.
- Now get_prediction take an execution object instead of an execution id.

### `back/main.py`
- Change endpoint `/dataset/upload/` signature, now it returns a Tuple with the experiment_id and the list of classes related to the task.
- The session_id parameter is now mandatory for the correct execution of the endpoints.
- Added endpoint `/dataset/` to see all the available datasets.
- Change endpoint `/dataset/upload/`, now it only creates a new dataset in the db
- Added endpoint `/experiment/create/{dataset_id}`, it works same as the old `/dataset/upload/` endpoint, but now it receives a dataset_id.


cristian-tamblay (changes requested):
DashAI/back/Database/datasets/3.json is very large (pypi allows 50-60MB max). What we can do is that on the first run of the dashai command the user is asked if he wants to download the default datasets (or maybe even in the GUI this could be done). You can also ask if you want to download the small (~kbs) medium (50-500MB) or large (1-10GBs) examples in order to have freedom in the size of the datasets. Better to get them out of the commit for the time being

DashAI/back/main.py
```diff
@@ -26,40 +30,206 @@
...
 
 @app.get("/dataset/task_name/{session_id}")
 async def get_task_name(session_id: int):
-    session_id = 0 # TODO Get session_id from user
-    return session_info[session_id]["task_name"]
+    """
+    Returns the task_name associated with the experiment of id session_id.
+    """
+    experiment : models.Experiment = db.session.query(models.Experiment).get(session_id)
+    main_task = Task.load(experiment.task_filepath)
```
cristian-tamblay: image
New DashAI installation crashes here after trying to upload a file

DashAI/back/dataloaders/classes/dataloader.py
```diff
@@ -0,0 +1,70 @@
+from abc import abstractmethod
+
+from datasets import ClassLabel
+
+from DashAI.back.config_object import ConfigObject
+
+
+class DataLoader(ConfigObject):
```
pbadillatorrealba: I would suggest renaming `DataLoader` to `BaseDataLoader`. 
I think it is a good pattern to indicate that this class is the base that others should implement. 
In general, I think it would be quite practical to use this in the rest of the classes in the package.

Source: 
- Scikitlearn base module: https://scikit-learn.org/stable/modules/classes.html#module-sklearn.base

DashAI/back/dataloaders/classes/dataloader.py
```diff
@@ -0,0 +1,70 @@
...
+    """
+    Abstract class with base methods for all data loaders
+    """
+
+    @abstractmethod
+    def load_data(self, dataset_path, url=None):
+        pass
```
pbadillatorrealba: All abstract methods must raise NotImplementedError as a default behavior.

> _exception `NotImplementedError` [...] In user defined base classes, abstract methods should raise this exception when they require derived classes to override the method, or while the class is being developed to indicate that the real implementation still needs to be added._

Source:
- https://docs.python.org/3/library/exceptions.html#NotImplementedError

```suggestion
        raise NotImplementedError
```

DashAI/back/dataloaders/classes/dataloader.py
```diff
@@ -0,0 +1,70 @@
...
+
+    @abstractmethod
+    def set_task_format(self, task):
+        pass
```
pbadillatorrealba:
```suggestion
        raise NotImplementedError
```

DashAI/back/dataloaders/classes/dataloader.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def split_dataset(self, dataset, params, class_column):
+        """
+        Split the dataset in train, test and validation data
+        """
```
pbadillatorrealba: As a public method, this function should have good documentation and hopefully examples and typing.

```suggestion
    def split_dataset(self, dataset: TODO, params: dict[str, Any], class_column: str) -> TODO:
        """Split the dataset in train, test and validation data

        Parameters
        ----------
        dataset : TODO
            ...
        params : dict[str, Any]
            Dict of parameters that configure the splitting process.
            These can be: 
            - stratify: bool = True indicates that the split will be stratified. 
            - test_size: float = TODO
            - val_size: float = TODO
            By default TODO
        class_column : str
            In the case of stratifying, indicate the column with which to stratify.

        Returns
        -------
        TODO
            The splitted dataset.
        """
```

pbadillatorrealba: In addition, I have some conceptual questions about this function: 

1. Why the parameters are passed in params and class_column (which is also parameter) independent of the dictionary. Moreover, is it necessary to have the parameters in params? Why you simply don't receive multiple parameters (something like def split_dataset(dataset, train_size, val_size, stratify, class_column, ...). In this case, you can make explicit which parameters the function receives (something that is not intuitive using a dictionary) and the types of each parameter.
2. I think it would be more correct that the function will make explicit that the operation sought is to generate three partitions: train test split. what if the user does not want to generate val and only use train and test?
3. What happens with the datasets that come previously separated?

DashAI/back/dataloaders/classes/dataloader.py
```diff
@@ -0,0 +1,70 @@
...
+        stratify_column = None
```
pbadillatorrealba: In general, there should be type and value checks on public function inputs.

```suggestion
        # input checks
        if not isinstance(params, dict):
            raise TypeError(f"params should be a string, got {type(class_column)}")
        if not isinstance(class_column, str):
            raise TypeError(
                f"stratify_column should be a string, got {type(class_column)}"
            )
        # as well the parameters in the params dict
        if not isinstance(params["stratify"], str):
            raise TypeError(
                f"params[\"stratify\"] should be a string, got {type(params['stratify'])}"
            )
        if not isinstance(params["stratify"], str):
            raise ValueError(f'params["stratify"] should be column in the dataset.')

        stratify_column = None
```

DashAI/back/dataloaders/classes/dataloader.py
```diff
@@ -0,0 +1,70 @@
...
+        if params["stratify"]:
+            stratify_column = class_column
```
pbadillatorrealba: - What if the dataset problem is not a classification problem and I still pass a column name?
- What if params does not have the ''stratify" key (it would raise an exception). Here you have to check first that the key exists.
- What if I don't check that the column exists in the dataset. In that case it would still be good to raise an exception (see checks above).

DashAI/back/dataloaders/classes/dataloader.py
```diff
@@ -0,0 +1,70 @@
...
+        test_val = params["test_size"] + params["val_size"]
+        val_size = params["val_size"] / test_val
```
pbadillatorrealba: I think it would be good to explain and document more this operation and then, test it. Are there edge cases that potentially leave out data?

DashAI/back/dataloaders/classes/dataloader.py
```diff
@@ -0,0 +1,70 @@
...
+        )
+        test_valid_split = train_split["test"].train_test_split(
+            test_size=val_size,
+            shuffle=params["shuffle"],
+            seed=params["seed"],
+            stratify_by_column=stratify_column,
+        )
+        dataset["train"] = train_split["train"]
+        dataset["test"] = test_valid_split["train"]
+        dataset["validation"] = test_valid_split["test"]
```
pbadillatorrealba, line 141: Does this generate a view or a copy?
How memory-intensive is this operation?

DashAI/back/dataloaders/classes/dataloader.py
```diff
@@ -0,0 +1,70 @@
...
+    def set_classes(self, dataset, class_index):
+        """
+        Set class column (for tabular data)
+        """
+        for split in ["train", "test", "validation"]:
+            if split in dataset:
+                label = dataset[split].column_names[class_index]
+                new_features = dataset[split].features.copy()
+                new_features[label] = ClassLabel(names=list(set(dataset[split][label])))
+                dataset[split] = dataset[split].cast(new_features)
```
pbadillatorrealba: I believe this operation is inefficient. Can't the class be accessed directly instead of copying all the features (i.e., what happens for a dataset that weighs 10 gb?).

DashAI/back/dataloaders/classes/dataloader.py
```diff
@@ -0,0 +1,70 @@
...
+        return dataset, label
+
+    def select_features(self, dataset, selected_features):
```
pbadillatorrealba: Again, if the function is public, then document better and type in if possible what the parameters are.

```suggestion
    def select_features(self, dataset: Dataset, selected_features: list[str]) -> Dataset:
```

DashAI/back/dataloaders/classes/dataloader.py
```diff
@@ -0,0 +1,70 @@
...
+            shuffle=params["shuffle"],
+            seed=params["seed"],
+            stratify_by_column=stratify_column,
+        )
+        dataset["train"] = train_split["train"]
+        dataset["test"] = test_valid_split["train"]
+        dataset["validation"] = test_valid_split["test"]
+        return dataset
+
+    def set_classes(self, dataset, class_index):
```
pbadillatorrealba: If the function is public, then document better and type in if possible what the parameters are.
```suggestion
    def set_classes(self, dataset, class_index: int):
```

DashAI/back/dataloaders/classes/dataloader.py
```diff
@@ -0,0 +1,70 @@
...
+        """
+        Set class column (for tabular data)
+        """
+        for split in ["train", "test", "validation"]:
```
pbadillatorrealba: If the function is public, write checks to determine if the dataset is of the corresponding type and if class_index is int and if it exists within the dataset.

DashAI/back/dataloaders/classes/dataloader.py
pbadillatorrealba: This functionality should be part of the dataset?
I think if there is automatic selection it should be part of the model.

DashAI/back/dataloaders/classes/dataloader.py
```diff
@@ -0,0 +1,70 @@
...
+                new_features = dataset[split].features.copy()
+                new_features[label] = ClassLabel(names=list(set(dataset[split][label])))
+                dataset[split] = dataset[split].cast(new_features)
+        return dataset, label
+
+    def select_features(self, dataset, selected_features):
+        """
+        Remove the features not selected for the dataset (for tabular data)
+        """
+        for split in ["train", "test", "validation"]:
```
pbadillatorrealba: Check that dataset is Dataset, that selected_features is list, that all list elements are strings and that all features exist in the dataset.

DashAI/back/dataloaders/classes/csv_dataloader.py
```diff
@@ -0,0 +1,35 @@
...
+from DashAI.back.dataloaders.classes.dataloader import DataLoader
+
+
+class CSVDataLoader(DataLoader):
+    """
+    Data loader for tabular data in CSV files
+    """
+
+    def load_data(self, dataset_path, params, file=None, url=None):
+        separator = params["separator"]
```
pbadillatorrealba, line 64: What happens if I do not specify the separator or the key does not exist in the parameter. Where is it documented that it must be in the dictionary.

DashAI/back/dataloaders/classes/csv_dataloader.py
```diff
@@ -0,0 +1,35 @@
...
+
+from DashAI.back.dataloaders.classes.dataloader import DataLoader
+
+
+class CSVDataLoader(DataLoader):
+    """
+    Data loader for tabular data in CSV files
+    """
+
+    def load_data(self, dataset_path, params, file=None, url=None):
```
pbadillatorrealba: Public method needs documentation and hopefully typed documentation and examples of use.

pbadillatorrealba: Do the parameters need to be passed in params? why not simply as function arguments? 
Thus, they can be explicitly declared and even include default types and values.

DashAI/back/dataloaders/classes/audio_dataloader.py
```diff
@@ -0,0 +1,31 @@
...
+
+from DashAI.back.dataloaders.classes.dataloader import DataLoader
+
+
+class AudioDataLoader(DataLoader):
+    """
+    Data loader for data from audio files
+    """
+
+    def load_data(self, dataset_path, params, file=None, url=None):
```
pbadillatorrealba: Public method needs documentation and hopefully typed documentation and examples of use.

pbadillatorrealba: Do the parameters need to be passed in params? why not simply as function arguments?
Thus, they can be explicitly declared and even include default types and values.

DashAI/back/database/db.py
```diff
@@ -2,7 +2,7 @@
 from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy.orm import sessionmaker
 
-engine = create_engine("sqlite:///dashAI.sqlite")
+engine = create_engine("sqlite:///DashAI/back/database/DashAI.sqlite")
```
pbadillatorrealba: This should ideally be parameterized in a config.py file.
file (fastapi's recommendation - https://fastapi.tiangolo.com/advanced/settings/#settings-in-another-module)
In general, any configuration should be centralized there.

DashAI/back/dataloaders/classes/image_dataloader.py
```diff
@@ -0,0 +1,31 @@
...
+
+from DashAI.back.dataloaders.classes.dataloader import DataLoader
+
+
+class ImageDataLoader(DataLoader):
+    """
+    Data loader for data from image files
+    """
+
+    def load_data(self, dataset_path, params, file=None, url=None):
```
pbadillatorrealba: Documenting and typing.
Same question, why the parameters in params?

DashAI/back/main.py
```diff
@@ -31,116 +31,11 @@
...
-
 
-@app.get("/play/{session_id}/{execution_id}/{input}")
-async def generate_prediction(session_id: int, execution_id: int, input_data: str):
-    session_id = 0  # TODO Get session_id from user
-    execution_id = 0  # TODO Get execution_id from user
-    main_task = session_info[session_id]["task"]
-    return str(main_task.get_prediction(execution_id, input_data))
+app.include_router(datasets.router)
+app.include_router(experiments.router)
```
pbadillatorrealba, line 36: Great!

DashAI/back/main.py
```diff
@@ -31,116 +31,11 @@
...
 
 
 if __name__ == "__main__":
+    db.Base.metadata.create_all(db.engine)
```
pbadillatorrealba, line 40: What if the database has already been initialized?

DashAI/back/routers/datasets.py
```diff
@@ -0,0 +1,112 @@
...
+        folder_path = os.path.realpath(folder_path)
+        db_dataset = models.Dataset(params.dataset_name, params.task_name, folder_path)
+        db.session.add(db_dataset)
+        db.session.flush()
+        db.session.commit()
+
+        # TODO remove this
+        session_info.dataset = dataset
+        session_info.task_name = params.task_name
+        session_info.task = Task.createTask(params.task_name)
```
pbadillatorrealba, line 139: What happened to this TODO?

DashAI/back/routers/datasets.py
```diff
@@ -0,0 +1,112 @@
...
+        return {"message": "Couldn't connect with DB."}
+
+
+@router.delete("/dataset/")
+async def delete_dataset():
+    return {"message": "To be implemented"}
+
+
+@router.put("/dataset/")
+async def insert_dataset():
```
pbadillatorrealba:
```suggestion
async def update_dataset():
```

DashAI/back/routers/datasets.py
```diff
@@ -0,0 +1,112 @@
...
+                "dataset_path": db_dataset.path,
+            }
+    except exc.SQLAlchemyError:
+        return {"message": "Couldn't connect with DB."}
+    return available_datasets
+
+
+@router.post("/dataset/")
+async def upload_dataset(
+    params: str = Form(), url: str = Form(None), file: UploadFile = File(None)
```
pbadillatorrealba, line 63: Can a pydantic model be used directly here?

DashAI/back/routers/datasets.py
```diff
@@ -0,0 +1,112 @@
...
+        return model
+    except pydantic.ValidationError as e:
+        raise HTTPException(
+            detail=jsonable_encoder(e.errors()),
+            status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
+        ) from e
+
+
+@router.get("/dataset/")
+async def get_dataset():
```
pbadillatorrealba, line 44: What if I want only a specific dataset?

pbadillatorrealba: Here I would leave the door open to implement a get with queries.
https://fastapi.tiangolo.com/tutorial/query-params/

DashAI/back/routers/experiments.py
```diff
@@ -0,0 +1,30 @@
...
+    return main_task.experimentResults
+
+
+@router.delete("/experiment/")
+async def delete_experiment():
+    return {"message": "To be implemented"}
+
+
+@router.put("/experiment/")
+async def modify_experiment():
```
pbadillatorrealba:
```suggestion
async def update_experiment():
```

DashAI/back/routers/experiments.py
```diff
@@ -0,0 +1,30 @@
...
+
+@router.post("/experiment/")
+async def run_experiment():
+    main_task = session_info.task
+    main_task.run_experiments(session_info.dataset)
+    return {"message": "Experiment done"}
+
+
+@router.get("/experiment/")
+async def get_experiment():
```
pbadillatorrealba, line 18: Call with queries not implemented.
Maybe it would be good if for this module that everything raise a `NotImplementedError` or something like that.

DashAI/back/main.py
pbadillatorrealba: In fastapi's coockiecutter there are some interesting scripts that could solve the initialization:

https://github.com/tiangolo/full-stack-fastapi-postgresql

DashAI/back/dataloaders/classes/dataloader.py
pbadillatorrealba: Also, what happens for datasets that are not tabular? Why should they have this function.

I think it would be good to think about an intermediate class TabularDataset that inherits from Dataset and that CSVDataset, JSONDataset, ParqueDataset should extend with this kind of specialized methods for this kind of data.

pbadillatorrealba (changes requested):
I left several code comments, but I have some major questions: 

- Why are parameters passed using dictionaries with params. I think this is a bad practice since you don't know what parameters to pass to the function since it's all inside the dictionary, not in the function signature. Instead, if it were in the function signature, you could directly pass the parameters to each function, have default values visible to the developer and even type. I strongly suggest that this be changed. Now, if it is for the pydantic issue, I think it would be better to build the pydantic object inside the function and not directly parse the parameters.
- It would be nice to have an intermediate DataLoader specifically for tabular data and then extend it to CSV, JSON, etc... 
- I see that there is a lot of duplicated functionality (like extracting zips, etc...) between dataloaders. Maybe it would be better to refactor this functionality into a function in the DataLoader and then just use it.
- I think it needs a lot of documentation, as well as testing of the functionality. Hopefully also implement typing.

DashAI/back/routers/datasets.py
cristian-tamblay: We had queries, but they were useless, eg:
```
@app.get("/dataset/task_name/{session_id}")
async def get_task_name(session_id: int):
    session_id = 0 # TODO Get session_id from user
    return session_info[session_id]["task_name"]
```
I've removed all the queries to get a clean start. We need to define what calls are needed in the frontend. Meanwhile we return all datasets.

DashAI/back/routers/datasets.py
cristian-tamblay: We need to discuss this. This may be the skeleton for a future orchestrator in the form of a class.

DashAI/back/routers/experiments.py
cristian-tamblay: This module needs to be reconsidered. It has "placeholders" of how things were called before, mainly:
```
main_task.run_experiments(session_info.dataset)
```
and
```
main_task.experimentResults
```

DashAI/back/routers/datasets.py
cristian-tamblay: What syntax are you thinking about? The next line calls a pydantic model who raises an exception if the form validation fails
```
params = parse_params(params)
```

```
def parse_params(params):
    """
    Parse JSON from string to pydantic model
    """
    try:
        model = DatasetParams.parse_raw(params)
        return model
    except pydantic.ValidationError as e:
        raise HTTPException(
            detail=jsonable_encoder(e.errors()),
            status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
        ) from e
```

DashAI/back/main.py
cristian-tamblay: I can be wrong, but the create_all method only creates the tables if they dont exist on the database.
```
Create all tables stored in this metadata.

Conditional by default, will not attempt to recreate tables already present in the target database.
```

DashAI/back/dataloaders/classes/dataloader.py
Kazymila: I'll change the name, it makes a lot of sense.

DashAI/back/dataloaders/classes/dataloader.py
Kazymila: I didn't know about that, thanks for the correction. It'll be changed.

DashAI/back/dataloaders/classes/dataloader.py
Kazymila: That method is not necessary anymore, so I removed it. because now the validation that the dataset has the format for the task, is done by the task itself.

DashAI/back/routers/datasets.py
cristian-tamblay: Try params: DatasetParams

Kazymila: The problem is that cannot upload the pydantic model with a file in the same endpoint, for that is necessary this 'trick' that parse a string with the json of parameters follow the pydantic model 'DatasetParams'. Other solution is divide this endpoint in two, one for upload the parameters directly with the pydantic model and other for upload the file and process the dataset.

DashAI/back/dataloaders/classes/dataloader.py
Kazymila: You're right is necessary a good documentation, I'll leave all the documentation in this branch tomorrow. Now in the branch front/dataloaders are more documentation.

Answering the questions:
1- The parameters are passed in this dictionary because in the endpoint this is obtained directly from the json received, so the 'splits' key is passed directly. But it effectively doesn't make sense to have this and separate the class column. I agree that it is better to make the parameters for this method explicit, I will change it.

2- From what was discussed in meetings I assumed that the three partitions had to exist, but it was thought that I could choose not to have a split by giving it a size of 0. Now that I see it better, this is not working with a partition at 0, but I can correct it to let him do it.

3- If the splits are defined before, this method is not called. This depend of a parameter ('folder_split' ir now renamed as 'splits_in_folders') that is checked in the endpoint.

DashAI/back/dataloaders/classes/dataloader.py
Kazymila: I totally agree, I'll change this method and the set_classes method in a Intermediate class for tabular dataloaders, these two are only for tabular data, so is necessary. Also the set_classes method because in non tabular data the clases are defined before in folders, following the new format.

DashAI/back/routers/experiments.py
pbadillatorrealba: Ok, let's solve it in another pr.

DashAI/back/dataloaders/classes/dataloader.py
pbadillatorrealba: Thanks for the answers!

Regarding point 2, if you think that the behavior when giving 0 in some split is wrong, then at the beginning verify that the parameter is greater than 0 and in the case that it is not, raise an exception. 

```python
if not isinstance(val_size, float):
    raise TypeError("val_size muest be float, got {type(val_size)}") .

if val_size < 0 or val_size > 1:
    raise ValueError("val_size must be in the (0, 1) range (0 and 1 not included), got {val_size}")
```

What is important is that what you decide is well documented 😀

DashAI/back/dataloaders/classes/dataloader.py
pbadillatorrealba: Thanks!

DashAI/back/main.py
pbadillatorrealba: Ok, maybe we can add more logic in a next pr.

DashAI/back/routers/datasets.py
pbadillatorrealba: Ok, I didn't know, thanks!

pbadillatorrealba:
Don't forget to include the changes to the changes in the changelog!

DashAI/back/routers/datasets.py
cristian-tamblay: > Other solution is divide this endpoint in two

I don't like this approach because it brings more problems than solutions, e.g: if one endpoint fails we need to communicate that the other, etc

DashAI/back/dataloaders/classes/dataloader.py
Kazymila: Thanks for your advice! I understand, I will clarify it in the code.

DashAI/back/dataloaders/classes/dataloader.py
Kazymila: I will add this checks, I had not considered it before because I assumed that the dictionary would come with the correct data, Which in fact may not always be the case.

DashAI/back/dataloaders/classes/dataloader.py
Kazymila: - I considered that if it's not a classification problem, the 'stratify' parameter would be false by default, because in the front end this option should not be shown to the user, so takes the default value of pydantic model DatasetParams, which is not really clear at all, so I will define this default value by setting the explicit parameters as you said, instead of receiving the dictionary. For that if you pass a class column, nothing would be done with it.
- Indeed, it is necessary to check if it exists, but now I will change that to the explicits parameters.
- Before calling this function, it is validated that the format of the dataset is compatible with the task and there it is verified that the dataset contains the column that was indicated as a class column, which was not yet included in this branch, but I will update it and leave a comment about this.

DashAI/back/dataloaders/classes/dataloader.py
Kazymila: This is because the splitting are made twice. First the dataset is splitting into a train split and a test-validation split, this is because the validation split is taken like a portion of the test split of the first splitting. For that the size of the test split for the first splitting is the sum of the sizes of test and validation splits.

Then, the test and validation splits are defined in the second splitting, where now the train size of this splitting is the final test split and the result of the test split is actually the validation split. So that ‘val_size’ is the proportion of the validation split of the rest of the data in the resulting test split form first splitting process.

An example, if I have a  train size of 0.8, a test size of 0.1 and a validation size of 0.1, in the first process we split the data in the train data with the 80% of the data, and a test-validation data with the 20% remaining. Then in then second process we divide this 20% in a 10% of test data and 10% of validation data. 

This is all because Hugging Face datasets use a method called ‘train_test_split’ similar to sklearn method, and for do the validation split we need do this trick. As recommended in this [forum](https://discuss.huggingface.co/t/how-to-split-main-dataset-into-train-dev-test-as-datasetdict/1090).

A bad case for this, is when your have very few samples, cause some splits can be empty. But for this is checked in the validation of dataset format that do the task, that you have an appropriate number of samples.

DashAI/back/dataloaders/classes/dataloader.py
Kazymila: This assigns the data of the splits to the dataset, so it's not a copy because the dataset doesn't have this data yet.

DashAI/back/dataloaders/classes/dataloader.py
Kazymila: I see your point, I thought the same, but here we are not copying the data, we are making a copy of a special dictionary that contains the information of the features. That dictionary correspond to 'dataset[split].features’ where we take the information of the columns from the data of a split. 

This is more explained in this [documentation about features](https://huggingface.co/docs/datasets/about_dataset_features), where also explain what is the [ClassLabel](https://huggingface.co/docs/datasets/v2.10.0/en/package_reference/main_classes#datasets.ClassLabel), that is like a feature type for set a column as a class column in Hugging Face datasets.

For set a column as class column we need to cast this feature as a ClassLabel that is what is doing in this code, that follows this [tutorial](https://huggingface.co/docs/datasets/process#cast) for process the data.

DashAI/back/dataloaders/classes/csv_dataloader.py
Kazymila: This dictionary of parameters is defined by the pydantic model 'DatasetParams', where you have the specific parameters of the dataloaders, among which is the separator for the case of the csv dataloader.
```
class DataLoaderParams(BaseModel):
    separator: Optional[str] = ","
    data_key: Optional[str] = "data"
    # add other params of dataloaders
...
class DatasetParams(BaseModel):
    task_name: str
    data_loader: str
    dataset_name: str
    class_column: Union[int, str] = -1
    splits_in_folders: bool = False
    splits: SplitParams
    dataloader_params: DataLoaderParams
```
So in the endpoint to upload datasets the key 'dataloader_params' that contain these parameters is called to pass the parameters to the dataloader in 'params' argument.
```
dataset = dataloader.load_data(
            dataset_path=folder_path,
            params=params.dataloader_params.dict(),
            file=file,
            url=url,
        )
```
When the input string with the parameters is parsed to the pydantic model, automatically the parameters that defines the model that do not exist in the parameters input are added, so the diccionary of parameter always will have all the keys with default values, even when one is not present at the user's input.
For that it is assumed that the key exists, but I will add the check anyways because is necessary that exist.

DashAI/back/dataloaders/classes/csv_dataloader.py
Kazymila: Is necessary the parameters in the params dictionary, because if you pass as explicit arguments you cannot generalize the parameters for all data loaders in the endpoint. If it were explicit arguments you would have to pass specifically the parameter it needs each dataloader, and with the dictionary you can pass any parameter it needs and this is internally obtained from the dictionary, so each dataloader gets the parameter it needs without having to make extra conditionals for each dataloader in the endpoint. 

This is considering that the dataloader called changes depending of the user input, so the user input have to indicate one of this passing it as a string from frontend to backend in the parameters. For that we can have a dict with the name associate to a dataloader and their instances.

```
init_classes = {"CSVDataLoader": CSVDataLoader(), "JSONDataLoader": JSONDataLoader()}
```

So the dataloader will be for example "CSVDataLoader", then at the next code this dataloader is called.
```
dataloader = init_classes[params.dataloader] # params.dataloader = " CSVDataloader"
dataset = dataloader.load_data(
            dataset_path=folder_path,
            params=params.dataloader_params.dict(),
            file=file,
            url=url,
        )
```
Here is called the load method of the dataloader chosen by the user, so the parameters are also variable. I cannot pass separator parameter directly because if I have a JSON I don't receive this parameter and I need other parameters. So this was the solution that I found more convenient.

DashAI/back/dataloaders/classes/tabular_dataloader.py
```diff
@@ -0,0 +1,107 @@
...
+                    label = class_column
+                else:
+                    label = dataset[split].column_names[class_column]
+                new_features = dataset[split].features.copy()
+                new_features[label] = ClassLabel(names=list(set(dataset[split][label])))
+                dataset[split] = dataset[split].cast(new_features)
+        return dataset, label
+
+    def select_features(
+        self, dataset: DatasetDict, selected_features: list[str]
```
cristian-tamblay: list[str] raises TypeError, from typing import List then use List instead of list

DashAI/back/dataloaders/classes/csv_dataloader.py
```diff
@@ -0,0 +1,73 @@
...
+
+class CSVDataLoader(TabularDataLoader):
+    """
+    Data loader for tabular data in CSV files
+    """
+
+    def load_data(
+        self,
+        dataset_path: str,
+        params: dict[str, any],
```
cristian-tamblay: Use Dict from typing, dict raises TypeError

cristian-tamblay (approved):

DashAI/back/dataloaders/classes/dataloader.py
pbadillatorrealba: Ok!

DashAI/back/dataloaders/classes/dataloader.py
pbadillatorrealba: Ok!

DashAI/back/dataloaders/classes/csv_dataloader.py
pbadillatorrealba: Thanks.

DashAI/back/dataloaders/classes/dataloader.py
```diff
@@ -0,0 +1,139 @@
...
+        val_size: float,
+        seed: int = None,
+        shuffle: bool = True,
+        stratify: bool = False,
+        class_column: str = None,
+    ) -> DatasetDict:
+        """
+        Split the dataset in train, test and validation data.
+
+        Args:
```
pbadillatorrealba: Add the split process documentation to the function docstring.
@cristian-tamblay , I copied and pasted @Kazymila 's comment, but I guess it needs to be better adapted.

```suggestion
        Split the dataset in train, test and validation data.

        First the dataset is splitting into a train split and a test-validation split, this is because the validation split is taken like a portion of the test split of the first splitting. For that the size of the test split for the first splitting is the sum of the sizes of test and validation splits.

        Then, the test and validation splits are defined in the second splitting, where now the train size of this splitting is the final test split and the result of the test split is actually the validation split. So that ‘val_size’ is the proportion of the validation split of the rest of the data in the resulting test split form first splitting process.

        An example, if I have a train size of 0.8, a test size of 0.1 and a validation size of 0.1, in the first process we split the data in the train data with the 80% of the data, and a test-validation data with the 20% remaining. Then in then second process we divide this 20% in a 10% of test data and 10% of validation data.

        Args:
```

DashAI/back/dataloaders/classes/dataloader.py
```diff
@@ -0,0 +1,139 @@
...
+
+        # The splitting are made twice -------------------------------------------
+        # First the dataset is splitting into a train split and a test-val split,
+        # this is because the validation split is taken like a portion of
+        # the test split of the first splitting (test-val).
+
+        # Then, the test and validation splits are defined in the second splitting
+        # where now, the train size of this splitting is the final test split and
+        # the result of the test split is actually the validation split.
+        # -------------------------------------------------------------------------
```
pbadillatorrealba: Move to the docstring.
```suggestion

```

pbadillatorrealba (changes requested):
One more docstring change and then it is approved!

pbadillatorrealba (approved):

PR #12, opened 2022-12-23
Front/interface adjustments

maximi4321:
# Changes

## Components
- `Upload.jsx`: File upload by drag and drop is blocked when a file is already uploaded.
- `AddModels.jsx`: The form for adding models is reset after adding a model.
- `ModelsTable.jsx`: Added a button to remove a model in case the user needs to do so
- `Results.jsx`: 
  - Added a table to show a summary of the main results in case of training multiple models.
  - Interface now adapts to display multiple metrics.
  - Added tabs to view model-specific results when training multiple models.
  - Errors are now displayed for all parameters in the parameter configuration form.
  - Fixed the height of the card to accommodate the length of the parameter list and the multiple metrics
- `ParameterForm.jsx`: Added a tooltip to provide information about the parameters of the model being configured.

## Layouts
- `ExperimentPipeline.jsx`: The "Run Experiment" button now sends the parameters of all models.

## Styles
Added svg icons for buttons, made adjustments in heights and colors.

## Summary
- Fixed some bugs and incomplete functionalities.
- Added a tooltip to show information about parameters of a model.
- The interface can now work with multiple models (of different types).
- Users can now remove models they have added by mistake when setting up an experiment
- Multiple metrics of a model can be displayed in the results
- Added icons and adjustments in css.


cristian-tamblay (approved):
Quité un par de archivos que estaban ignorados pero seguian en el repo. Testie un flujo normal y funcionó.

PR #21, opened 2023-03-01
Back/db tables

cristian-tamblay:
# DB Tables
This PR intention is to implement new database model, hopefully giving all the tools necessary for the next pipeline orchestrator.

The model diagram is on DashAI's Miro board.
### `DashAI/back/database/models.py`
We are now using the modern version of SQLAlchemy Mapped syntax
- We have 4 tables now: Dataset, Experiment, Model and Run
- Dataset One-to-Many Experiment
- Experiment One-to-Many Model
- Model One-to-One Run
- Model and Run cascades on deletion.

### `DashAI/back/models/enums/states.py`
- Added states for the future state machine.

### DIFF
--- a/DashAI/back/database/models.py
+++ b/DashAI/back/database/models.py
@@ -1,10 +1,9 @@
-from sqlalchemy import Enum, String, ForeignKey, JSON
-from sqlalchemy.orm import Mapped
-from sqlalchemy.orm import relationship
-from sqlalchemy.orm import mapped_column
-from sqlalchemy.orm import DeclarativeBase
-from DashAI.back.models.enums.states import State
 from typing import List
+
+from sqlalchemy import JSON, Enum, ForeignKey, String
+from sqlalchemy.orm import DeclarativeBase, Mapped, mapped_column, relationship
+
+from DashAI.back.models.enums.states import State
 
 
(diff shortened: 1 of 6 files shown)


DashAI/back/database/models.py
```diff
@@ -1,73 +1,61 @@
...
-#     def get_results(self):
-#         results = {}
-#         for execution in self.executions:
-#             results[execution.id] = {
-#                 "train": execution.train_results,
-#                 "test": execution.test_results,
-#             }
-#         return results
 
+class Model(Base):
```
pbadillatorrealba: Maybe it would be a good idea to change the name to `ModelConfig` or `ModelInstance`, so that it is understood that this is a specific configuration of a particular model and not a table with the specification of each model.

pbadillatorrealba (commented):
I left a comment regarding the name of a table.
Question: what would happen if you run mypy to check the types (only the types of this pr 😪)?

DashAI/back/database/models.py
cristian-tamblay: Yes! I forgot to change that, thanks

pbadillatorrealba (approved):

PR #22, opened 2023-03-01
Added serving to FastAPI, removed external serving, fixed paths on front

cristian-tamblay:
# Serving from FastAPI
This PR intention is to remove a secondary server to serve our front end. Also, it enhances GitHub workflow to build the frontend and then run pytest.

Fixed some .gitignore redundancies

### `.github/workflows/build-test.yaml`
- Unified pytest and npm test to a single workflow with two jobs
- pytest job uses the react build generated on the npm cli cycle.

### `DashAI/__init__.py`
- Removed the secondary process that ran the server on python's http.server

### `DashAI/back/database/db.py`
- Fixed deprecated import in SQLAlchemy 2.0

### `DashAI/back/main.py`
- Enabled frontend serve and static files on FastAPI
- Renamed app path from / to /app
- Renamed api path from / to /api
- Added database health check on execution.

### Frontend
- Renamed a bunch of paths from images/* to /images/*
- Renamed NavBar's paths from /* to /app/*


.gitignore
```diff
@@ -162,7 +162,6 @@ cython_debug/
 .vscode
 dist
 DashAI.egg-info
-build
```
pbadillatorrealba, line 165: I don't think it's a good idea to get builds from the gitignore.
The idea is that the repository does not contain any build file, but that the build is created outside.

DashAI/__init__.py
```diff
@@ -7,29 +7,19 @@
...
 
 def run():
-    url = "http://localhost:3000/"
+    url = "http://localhost:8000/app"
 
-    procs = [
-        Popen(
+    p = Popen(
             "python -m DashAI.back.main",
             shell=True,
```
pbadillatorrealba: If two processes are no longer running, then there is no need to launch separate processes.
Here you would launch the FastAPI server directly.
```suggestion
from DashAI.back.main import app

if __name__ == "__main__":
    uvicorn.run(app, host="127.0.0.1", port=8000)
```

And delete everything else.

DashAI/back/main.py
```diff
@@ -1,42 +1,56 @@
...
-    allow_origins=origins,
-    allow_credentials=True,
-    allow_methods=["*"],
-    allow_headers=["*"],
-)
+@app.get("/")
+async def redirect():
+    return RedirectResponse(url="/app")
+
+
```
pbadillatorrealba: Perhaps for the backlog, any route that is not registered in any router should be returned by the application. 
So if for example I enter localhost /asddsd it does not deliver a bad request directly to the server but returns the app.

DashAI/back/main.py
```diff
@@ -1,42 +1,56 @@
...
+api = FastAPI(title="DashAI API")
+
+api.include_router(datasets.router)
+api.include_router(experiments.router)
+
+app.mount("/static", StaticFiles(directory="DashAI/front/build/static"), name="static")
+app.mount("/images", StaticFiles(directory="DashAI/front/build/images"), name="images")
+templates = Jinja2Templates(directory="DashAI/front/build")
```
pbadillatorrealba: Why use jinja if there are no templates to compile?

DashAI/front/.gitignore
```diff
@@ -8,9 +8,6 @@
 # testing
 /coverage
 
-# production
-/build
-
```
pbadillatorrealba, line 13: Same here, I don't think it's a good idea to pull builds from the gitignore.

pbadillatorrealba (changes requested):
Some comments on performance and other server details.

.gitignore
cristian-tamblay: This build was redundant because there is already a ignore of build on line 10: https://github.com/DashAISoftware/DashAI/blob/c2da386cf4894fa8cc309f0c3ac4eb582a9e3f94/.gitignore#L10

DashAI/front/.gitignore
cristian-tamblay: The build/ from root gitignore catches this one, it was redundant

DashAI/back/main.py
cristian-tamblay: I was following [this](https://stackoverflow.com/a/64147628) answer on stackoverflow.
You are correct. There was an easier way
```
@app.get("/app/{full_path:path}")
async def read_index():
    return FileResponse('DashAI/front/build/index.html')
```

DashAI/__init__.py
cristian-tamblay: Done, i also added the auto pop up from the browser after 1s

DashAI/back/main.py
cristian-tamblay: Now localhost/asddsd and localhost/ will redirect to /app/
BUT localhost/app/asddsd should be handled by React routing and not by the backend endpoint.

DashAI/back/main.py
cristian-tamblay: We also now load statics without the app.mount black box and redirects if the file is not found.
```
@app.get("/{full_path:path}")
async def serve_files(full_path: str):
    try:
        if full_path == "":
            return RedirectResponse(url="/app/")
        path = f"DashAI/front/build/{full_path}"
        os.stat(path)  # This checks if the file exists
        return FileResponse(path)  # You can't catch the exception here
    except FileNotFoundError:
        return RedirectResponse(url="/app/")
```

pbadillatorrealba: Nice!

pbadillatorrealba (approved):

PR #24, opened 2023-03-06
Merged ModelInstance and Run tables

cristian-tamblay:
### `DashAI/__init__.py`
- Base was refactored to database/models, so fixed import in __init__

### `DashAI/back/database/models.py`
- Merged the ModelInstance table with the Run table


DashAI/back/database/models.py
```diff
@@ -43,21 +43,8 @@ class ModelInstance(Base):
...
-    """
-    Table to store all the information about a specific run of a model.
-    """
-    id: Mapped[int] = mapped_column(primary_key=True)
-    model_id: Mapped[int] = mapped_column(ForeignKey("model_instance.id"))
     train_results: Mapped[JSON] = mapped_column(JSON)
     test_results: Mapped[JSON] = mapped_column(JSON)
     validation_restuls: Mapped[JSON] = mapped_column(JSON)
-    weights_path: Mapped[str] = mapped_column(String)
+    binary_path: Mapped[str] = mapped_column(String)
```
pbadillatorrealba: Two things here:

1. I've seen a lot of the artifact concept (https://h2o.ai/wiki/artifacts/, https://machine-learning.paperspace.com/wiki/artifacts , etc.). Maybe it would be good to change the name to fit that standard?
```suggestion
    artifact_path: Mapped[str] = mapped_column(String)
```
2. Where would other training outputs such as interpretability graphs or roc curves graphs be stored?

cristian-tamblay: 1. Name makes sense.
2. Thats a good question, i would leave those graphs as part of a key (exploiting the fact JSONs don't have a fixed length) of the train_results/test_results/validation_results JSON's. I have an idea but thats part of the metric object which isnt implemented. The JSON should contain: {"roc_curve": path_to_curve_png ... } or maybe { "roc_curve" : base64_encoded_png ... } (but i dislike second option)

DashAI/back/database/models.py
```diff
@@ -43,21 +43,8 @@ class ModelInstance(Base):
...
-    model_id: Mapped[int] = mapped_column(ForeignKey("model_instance.id"))
     train_results: Mapped[JSON] = mapped_column(JSON)
     test_results: Mapped[JSON] = mapped_column(JSON)
     validation_restuls: Mapped[JSON] = mapped_column(JSON)
-    weights_path: Mapped[str] = mapped_column(String)
+    artifact_path: Mapped[str] = mapped_column(String)
     trained: Mapped[Boolean] = mapped_column(Boolean)
-    model: Mapped["ModelInstance"] = relationship(
-        back_populates="model_instance", cascade="all, delete"
-    )
```
pbadillatorrealba: Responding to what you commented for the artifact comments:

I was looking at how mflow implements this. In its most basic form (only with folders) they keep 4 different folders for each run: artifacts, metrics, params and tags. 
Inside artifacts live the serialized model as well as the graphics, inside metrics live all the metrics, params live the parameters (which would be ModelInstance in our case) and tags like the start and end date of execution, user that launched the run, execution environment, etc...

Maybe we could adapt this structure to more structured dash ideas: 

- artifacts to be a json with key the artifact name and value the path.
- metrics leave it as it is
- add tags to the same run
- parameters: merge model instance with this table.

Below is a suggestion of the above (deleting ModelInstance table):

```python
class RunStatus:
    NOTSTARTED = 0
    STARTED = 1
    FINISHED  = 2
    ERROR = 3

class Run(Base):
    __tablename__ = "run"
    """
    Table to store all the information about a specific run of a model.
    """
    id: Mapped[int] = mapped_column(primary_key=True)
    # model and parameters
    model_name: Mapped[str] = mapped_column(String)
    parameters: Mapped[JSON] = mapped_column(JSON)
    # metrics
    train_metrics: Mapped[JSON] = mapped_column(JSON)
    test_metrics: Mapped[JSON] = mapped_column(JSON)
    validation_metrics: Mapped[JSON] = mapped_column(JSON)
    # artifacts
    artifacts: Mapped[str] = mapped_column(JSON)
    # metadata
    run_name: Mapped[str] = mapped_column(String)
    run_description: Mapped[str] = mapped_column(String)
    status: Mapped[Enum] = mapped_column(Enum(RunStatus), nullable=False)
    start_time: Mapped[DateTime] = mapped_column(DateTime)
    end_time: Mapped[DateTime] = mapped_column(DateTime)
    end_time: Mapped[DateTime] = mapped_column(DateTime)
```

We could still discuss this at a meeting, what do you think?

cristian-tamblay: This is a good idea, i made some minor fixes (its missing experiment_id, end_time is duplicated, etc). Hope this has enough info for our future orchestrator

pbadillatorrealba (approved):

PR #27, opened 2023-03-14
Implemented all datasets endpoints

cristian-tamblay:
# Dataset endpoints
New dataset endpoints on `DashAI/back/api_v1` follow the RESTful api paradigm

## Changes

- Implemented put and delete on datasets
- Implemented get on /dataset/ and /dataset/id
- Now all datasets raise HTTPException if an error occurs.
- Fixed docstring styling
- Added info on CHANGELOG, removed dates so its easier to add lastest changes in upcoming PRs

### DIFF
--- a/DashAI/back/api/api_v1/endpoints/datasets.py
+++ b/DashAI/back/api/api_v1/endpoints/datasets.py
@@ -14,10 +14,6 @@ from DashAI.back.core.config import model_registry, settings
 from DashAI.back.database import models
 from DashAI.back.database.db import session
 from DashAI.back.dataloaders.classes.csv_dataloader import CSVDataLoader
-
-# from Dataloaders.classes.audioDataLoader import AudioDataLoader
-# from Dataloaders.classes.csvDataLoader import CSVDataLoader
-# from Dataloaders.classes.imageDataLoader import ImageDataLoader
 from DashAI.back.dataloaders.classes.dataloader_params import DatasetParams
 from DashAI.back.dataloaders.classes.json_dataloader import JSONDataLoader
 from DashAI.back.tasks import (
@@ -32,7 +28,7 @@ log = logging.getLogger(__name__)
 router = APIRouter()
 
 # TODO: Do this in a better way, this was the fastest fix to get str -> class
-# Maybe each class should have its respective representative string as variable
+# and start testing. We should do this mapping on Dataloaders Registry.
 
 dataloaders = {"CSVDataLoader": CSVDataLoader(), "JSONDataLoader": JSONDataLoader()}
 tasks = {
@@ -49,12 +45,12 @@ def parse_params(params):
     Parameters
     ----------
     params : str
-        JSON with parameters for load the data in a string.
+        Stringified JSON with parameters.
 
     Returns
     -------
     BaseModel
-        Pydantic model parsed from JSON in parameters string.
+        Pydantic model parsed from Stringified JSON.
 
     Raises
     ------
--- a/DashAI/back/api/api_v1/endpoints/datasets.py
+++ b/DashAI/back/api/api_v1/endpoints/datasets.py
@@ -4,13 +4,13 @@ import shutil
 from typing import Union
 
 import pydantic
-from fastapi import APIRouter, File, Form, UploadFile, status
+from fastapi import APIRouter, File, Form, Response, UploadFile, status
 from fastapi.encoders import jsonable_encoder
 from fastapi.exceptions import HTTPException
 from sqlalchemy import exc
 
 from DashAI.back.api.api_v0.endpoints.session_class import session_info
-from DashAI.back.core.config import model_registry, settings
+from DashAI.back.core.config import settings
 from DashAI.back.database import models
 from DashAI.back.database.db import session
 from DashAI.back.dataloaders.classes.csv_dataloader import CSVDataLoader
@@ -51,11 +51,6 @@ def parse_params(params):
     -------
     BaseModel
         Pydantic model parsed from Stringified JSON.
-
-    Raises
-    ------
-    ValidationError
-        If params can't be validated by the pydantic model
     """
     try:
         model = DatasetParams.parse_raw(params)
@@ -75,13 +70,8 @@ async def get_datasets():
 
     Returns
     -------
-    List[Dict]
-        List of datasets
-
-    Raises
-    ------
-    SQLAlchemyError
-        If the database connection or query fails.
+    List[JSON]
+        List of dataset JSONs
     """
 
     try:
@@ -111,13 +101,8 @@ async def get_dataset(dataset_id: int):
 
     Returns
     -------
-    Dict
-        Dataset dict with the specified id.
-
-    Raises
-    ------
-    SQLAlchemyError
-        If the database connection or query fails.
+    JSON
+        JSON with the specified dataset id.
     """
     try:
         dataset = session.query(models.Dataset).get(dataset_id)
@@ -134,7 +119,7 @@ async def get_dataset(dataset_id: int):
     return dataset
 
 
-@router.post("/")
+@router.post("/", status_code=status.HTTP_201_CREATED)
 async def upload_dataset(
     params: str = Form(), url: str = Form(None), file: UploadFile = File(None)
 ):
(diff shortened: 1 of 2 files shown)


DashAI/back/api/api_v1/endpoints/datasets.py
```diff
@@ -9,7 +11,8 @@
 
 from DashAI.back.api.api_v0.endpoints.session_class import session_info
 from DashAI.back.core.config import model_registry, settings
-from DashAI.back.database import db, models
+from DashAI.back.database import models
+from DashAI.back.database.db import session
 from DashAI.back.dataloaders.classes.csv_dataloader import CSVDataLoader
 
 # from Dataloaders.classes.audioDataLoader import AudioDataLoader
```
Riul1999: Are these comments there for future implementation?

DashAI/back/api/api_v1/endpoints/datasets.py
```diff
@@ -40,11 +46,20 @@ def parse_params(params):
     """
     Parse JSON from string to pydantic model
 
-    Args:
-        params (str): JSON with parameters for load the data in a string.
+    Parameters
+    ----------
+    params : str
```
Riul1999, line 47: params is a string or a JSON?

Riul1999 (changes requested):

DashAI/back/api/api_v1/endpoints/datasets.py
cristian-tamblay: It's both! Input is of the type str, but its a stringified JSON

DashAI/back/api/api_v1/endpoints/datasets.py
cristian-tamblay: Future dataset endpoint should have all dataloaders options. It should be part of the dataloader registry imports instead. Removed

DashAI/back/api/api_v1/endpoints/datasets.py
```diff
@@ -161,14 +237,88 @@ async def upload_dataset(
...
+        if not dataset:
+            raise HTTPException(
+                status_code=status.HTTP_404_NOT_FOUND, detail="Dataset not found"
+            )
+        session.delete(dataset)
+        shutil.rmtree(dataset.file_path, ignore_errors=True)
+        session.commit()
+        return {"ok": True}
+    except (exc.SQLAlchemyError, OSError) as e:
+        log.error(e)
```
pbadillatorrealba:
```suggestion
        log.exception(e)
```

DashAI/back/api/api_v1/endpoints/datasets.py
```diff
@@ -161,14 +237,88 @@ async def upload_dataset(
         return model_registry.task_to_components(params.task_name)
```
cristian-tamblay: @pbadillatorrealba @Riul1999  I think this is not necessary and we should have another endpoint returning the models list given a task. This is forcing the flow of the frontend to start experiment right after uploading dataset. What do you think? 
My claim is this endpoint should only return the created object or a message like {"ok":true}

DashAI/back/api/api_v1/endpoints/datasets.py
```diff
@@ -58,47 +64,89 @@ def parse_params(params):
...
+    """
+
+    try:
+        all_datasets = session.query(models.Dataset).all()
+        if not all_datasets:
+            raise HTTPException(
+                status_code=status.HTTP_404_NOT_FOUND, detail="No datasets found"
+            )
+    except exc.SQLAlchemyError as e:
+        log.error(e)
```
pbadillatorrealba: I believe that here it would be better to put an Exception rather than an error, since it is more serious than an error the fact that it is not possible to connect to the database.
```suggestion
        log.exception(e)
```

Additional question: why not use logger instead of log?

DashAI/back/api/api_v1/endpoints/datasets.py
```diff
@@ -107,7 +155,10 @@ async def upload_dataset(
         os.makedirs(folder_path)
     except FileExistsError as e:
         log.error(e)
-        return {"message": "Dataset name already exists."}
+        raise HTTPException(
+            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
```
pbadillatorrealba: The 500 errors are intended for critical server errors. In this case, it is a request problem, so it would return a 409 Conflict.

Reference: https://stackoverflow.com/questions/3825990/http-response-code-for-post-when-resource-already-exists

DashAI/back/api/api_v1/endpoints/datasets.py
pbadillatorrealba: In which case this function could raise an OSError? Is it correct that both exceptions are treated together?

DashAI/back/api/api_v1/endpoints/datasets.py
```diff
@@ -144,31 +195,98 @@ async def upload_dataset(
...
+        if name or task_name:
+            session.commit()
+            session.refresh(dataset)
+            return dataset
+        else:
+            raise HTTPException(
+                status_code=status.HTTP_304_NOT_MODIFIED, detail="Record not modified"
+            )
+    except exc.SQLAlchemyError as e:
+        log.error(e)
```
pbadillatorrealba:
```suggestion
        log.exception(e)
```

DashAI/back/api/api_v1/endpoints/datasets.py
```diff
@@ -58,47 +64,89 @@ def parse_params(params):
...
-                "dataset_task_name": db_dataset.task_name,
-                "dataset_path": db_dataset.path,
-            }
+        dataset = session.query(models.Dataset).get(dataset_id)
+        if not dataset:
+            raise HTTPException(
+                status_code=status.HTTP_404_NOT_FOUND, detail="Dataset not found"
+            )
     except exc.SQLAlchemyError as e:
         log.error(e)
```
pbadillatorrealba:
```suggestion
        log.exception(e)
```

DashAI/back/api/api_v1/endpoints/datasets.py
```diff
@@ -58,47 +64,89 @@ def parse_params(params):
...
+    List[JSON]
+        List of dataset JSONs
+    """
+
+    try:
+        all_datasets = session.query(models.Dataset).all()
+        if not all_datasets:
+            raise HTTPException(
+                status_code=status.HTTP_404_NOT_FOUND, detail="No datasets found"
+            )
```
pbadillatorrealba, line 82: What if you simply return an empty array?

pbadillatorrealba (changes requested):
Some small logging comments and return error codes.

DashAI/back/api/api_v1/endpoints/datasets.py
cristian-tamblay: There are [mixed](https://stackoverflow.com/questions/11746894/what-is-the-proper-rest-response-code-for-a-valid-request-but-an-empty-data) opinions on this topic. But most comments are in favour of 404 when you query an unexisting object

DashAI/back/api/api_v1/endpoints/datasets.py
cristian-tamblay: OSError raises when [shutil.rmtree fails](https://stackoverflow.com/questions/5962152/what-is-a-complete-list-of-exceptions-that-can-be-thrown-by-shutil-rmtree). I've separated them to have different details

DashAI/back/api/api_v1/endpoints/datasets.py
pbadillatorrealba: Maybe I'll stick more with the first answer.
But ok, let's leave it as it is.

DashAI/back/api/api_v1/endpoints/datasets.py
pbadillatorrealba: Nice!

DashAI/back/api/api_v1/endpoints/datasets.py
pbadillatorrealba: I also support this, so it doesn't break with the concept that this API is only for managing datasets and nothing else.

pbadillatorrealba (approved):

PR #28, opened 2023-03-16
Front/debugging

maximi4321:
# Changes

- Resolved warnings related to handling `undefined` values.
- Made some modifications to the `ParameterForm.jsx` component to handle two warnings.
  - One related to having a `useEffect` after an `if` with `return`.
  - Another related to the default values defined in `defaultProps`.
- Moved a block of code in `Upload.jsx` to fix a warning related to defining functions after an `if` with `return`.


cristian-tamblay (approved):
Loading tabs bugs are fixed.

pbadillatorrealba (approved):

PR #29, opened 2023-03-21
Back/test/dataset endpoints

cristian-tamblay:
# Dataset endpoint testing
Tests for new dataset endpoint

## Changes

- Added tests for all /dataset endpoints
- Fixed some docstrings for more clarity
- Removed dates from changelog for easier PRs
- Fixed DB session file name
- DB session now change depending on test/production


DashAI/back/core/config.py
```diff
@@ -28,6 +28,7 @@
 
 class Settings(BaseSettings):
     DB_PATH: str = "DashAI/back/database/DashAI.sqlite"
+    TEST_DB_PATH: str = "tests/back/database/test.sqlite"
```
pbadillatorrealba: Maybe this constant should be called `TABULAR_TEST_DATASET_PATH`
The other thing: this should be here and not in the same file where it is used for test?

tests/back/database/test_sql_app.py
```diff
@@ -0,0 +1,90 @@
...
+    try:
+        db = TestingSessionLocal()
+        yield db
+    finally:
+        db.close()
+
+
+api_v1.dependency_overrides[get_db] = override_get_db
+
+client = TestClient(app)
```
pbadillatorrealba: Perhaps later on it would be convenient to have this as a conftest, so as not to duplicate so much of this code while centralizing the test configurations.

pbadillatorrealba (changes requested):
Leave a small comment in the code.
Anyway I think two changes need to be made:

1. Name that now the connection to the database is parameterized somewhere (maybe in the pr and in the testing section in the development of the readme).
2. Change the path of the implemented tests: according to what I saw, the tests are focused on the dataset API, not on the database. Therefore, I would move them to: `tests/back/api/test_datasets.py` or something like that.

DashAI/back/core/config.py
cristian-tamblay: I think its better on the same file as test, no idea what i was thinking. But it is indeed a database that can hold any task, so i will mantain TEST_DB_PATH name

pbadillatorrealba (approved):

PR #30, opened 2023-03-21
dashai command added as entry point in setup file

inunezn7:
The executable file "dashai" was deleted and "dashai" was added as entry point instead. This entry point executes the run function inside DashAI package.

Other changes:
    - pytorch version in requirements and "--find-links" pytorch line in extra_requirementes in setup.py file. They generated an error when running the software probably due to the MacOS and CUDA.
    - Uncommented ".python-version" in .gitignore
    - .DS_Store added to .gitignore for MacOS

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -84,7 +84,7 @@ ipython_config.py
 # pyenv
 #   For a library or package, you might want to ignore these files since the code is
 #   intended to run in multiple environments; otherwise, check them in:
-# .python-version
+.python-version
 
 # pipenv
 #   According to pypa/pipenv#598, it is recommended to include Pipfile.lock in version control.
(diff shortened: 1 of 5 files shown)


pbadillatorrealba:
I tried to test it locally and it failed. I don't know if something is missing?
How did you test it?

cristian-tamblay:
It works for me after some changes:
Removed torch from setup. The syntax is incorrect (this had the same error from [#31](https://github.com/DashAISoftware/DashAI/pull/31#issuecomment-1478264308))
There was an `__init__.py` missing on `DashAI/back/models/scikit_learn/`
Then
```
pip install .
dashai
```
Works for me

Added these changes.

pbadillatorrealba:
Fantastic, now it works fine for me!
Could you add this information to the evelopment section of the readme?

PR #31, opened 2023-03-21
Back/cache workflow

cristian-tamblay:
# New cache for github workflow
This cache reduces 50% time on github's python installation

## Changes

- Implementented a python env cache, that is bypassed only when modifying setup.py, requirements or requirements-dev

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -38,8 +38,8 @@ jobs:
           path: ${{ env.pythonLocation }}
           key: ${{ env.pythonLocation }}-${{ hashFiles('setup.py') }}-${{ hashFiles('requirements.txt') }}-${{ hashFiles('requirements-dev.txt') }}
       
-      - run: pip install -upgrade --upgrade-strategy eager -r requirements.txt -e .
-      - run: pip install -upgrade --upgrade-strategy eager -r requirements-dev.txt -e .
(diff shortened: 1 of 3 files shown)


cristian-tamblay:
There is a problem with extras_require and pip install -e, if anyone has an idea how to fix it.
image

This improved pip install time from 6 minutes to 3minutes when using cache

.github/workflows/build-test.yaml
```diff
@@ -32,9 +32,14 @@ jobs:
...
         with:
           python-version: '3.10'
-          cache: 'pip'
-      - run: pip install -r requirements.txt
-      - run: pip install -r requirements-dev.txt
+        
+      - uses: actions/cache@v3
+        with:
+          path: ${{ env.pythonLocation }}
+          key: ${{ env.pythonLocation }}-${{ hashFiles('setup.py') }}-${{ hashFiles('requirements.txt') }}-${{ hashFiles('requirements-dev.txt') }}
```
pbadillatorrealba, line 39: If I understand correctly, this generates a directory for each different version of setup.py, requirements.txt and requirements-dev.txt?


What happens with outdated environments, are they deleted after a while?

cristian-tamblay: image

pbadillatorrealba: Nice!

pbadillatorrealba (approved):

PR #32, opened 2023-03-23
Front/material UI

maximi4321:
# Changes
- Installed material-ui.
- Installed [craco](https://www.npmjs.com/package/@craco/craco) in dev dependencies.
- Added `craco.config.js` file to use styled-components as the style engine for material-ui.
- modified `package.json` to use craco instead of react-scripts.
- Modified `index.js` to use material-ui ThemeProvider instead of the styled-components version.
- Replaced a Bootstrap dropdown with a material-ui dropdown in `ParameterForm` for testing purposes.


DashAI/front/src/App.test.jsx
```diff
@@ -2,15 +2,17 @@ import React from 'react';
...
 import App from './App';
 import theme from './styles/theme';
 
+const myTheme = createTheme(theme);
+
 test('loads and displays model', async () => {
   render(
     <Router>
-      <ThemeProvider theme={theme}>
+      <ThemeProvider theme={myTheme}>
```
pbadillatorrealba, line 18: It seems to me that besides wrapping the app with the theme, it is also a good idea to add X, since it standardizes the default css according to MUI styles.

```suggestion
      <ThemeProvider theme={myTheme}>
          <CssBaseline />
```

Source: 
- https://github.com/mui/material-ui/blob/master/examples/material-cra/src/index.js
- https://mui.com/material-ui/react-css-baseline/

pbadillatorrealba (commented):
Small comment on a MUI component.

DashAI/front/src/App.test.jsx
maximi4321: Thank you!
I added the CssBaseline component along with a new MUI global style because some CSS rules were being overwritten.

pbadillatorrealba (approved):

PR #36, opened 2023-04-02
Update Experiment table to MUI

pbadillatorrealba:
This pr updates the table of experiments to the MUI components.

The changes are: 
- The experiment table and new experiment button is now implemented using MUI's Datagrid and Button.
- A title and a container (Paper) are added to the experiment table.
- The component is completely defined in its file (before it was between main.jsx and experimentstable.jsx).

Note: this pr depends of the approvation of #35

PR #37, opened 2023-04-05
Separate experiment API from experiment table.

pbadillatorrealba:
This pr has as main focus to separate the experiments APIs (CRUD) from the experiments table component. Changes: 

- Moved the experiments API to its own module in `src/api/experiments`. Is by default typed.

- ExperimentsTable component now imports these methods and executes them in an asynchronous wrapper inside the component.

- Interactivity is added through snackbars/toast and loading animations. These can be seen when the getExperiments request fails and when an experiment is deleted (correctly or incorrectly).

- Typescript is enabled on the front end.

- Interfaces are declared for the main application objects: Experiment, Dataset and Run. These can be found in the `src/types` folder.

If this change is approved, it is expected that the rest of the APIs will be declared in the same or similar way as this one; that is, a module with the API declaration and the component imports the functions that it requires.

Note: this pr depends on #36


DashAI/front/src/api/experiment.ts
cristian-tamblay, line 1: If we are switching to typescript on some files, shouldn't we change all the jsx files to ts? What do you think about mixing javascript and typescript?

DashAI/front/src/api/experiment.ts
```diff
@@ -0,0 +1,33 @@
...
+}: {
+  id: string;
+  formData: object;
+}): Promise<object> => {
+  const response = await api.patch(`experiment/${id}`, formData);
+  return response.data;
+};
+
+export const deleteExperiment = async (id: string): Promise<object> => {
+  const response = await api.delete(`experiment/${id}`);
```
cristian-tamblay: I think we are missing the path here. We also have a .env defining api paths, maybe we should use them here.
```suggestion
  const response = await api.delete(`/api_v1/experiment/${id}`);
```

DashAI/front/src/api/experiment.ts
```diff
@@ -0,0 +1,33 @@
...
+};
+
+export const updateExperiment = async ({
+  id,
+  formData,
+}: {
+  id: string;
+  formData: object;
+}): Promise<object> => {
+  const response = await api.patch(`experiment/${id}`, formData);
```
cristian-tamblay:
```suggestion
  const response = await api.patch(`/api_v1/experiment/${id}`, formData);
```

cristian-tamblay (changes requested):

DashAI/front/src/api/experiment.ts
pbadillatorrealba: As far as I understand it is not necessary. Just type the ts files and their dependencies. 
It appears that changes may be gradual.
In this case it was only necessary to type the interfaces and the API file.

cristian-tamblay (approved):

PR #39, opened 2023-04-10
Poc/plugin system

inunezn7:
First implementation of the plugin system
- The software's core imports installed plugins that are declared as entry points in their config file (pyproject.toml). Inside the folder named "plugins" are some examples of that for tabular classification.
- An entry point was added to update the registry system with the installed plugins at runtime. The same function of the entry point is executed when DashAI is launched.
- For allowing devs to work in the DashAI components normally, this is, without having to install the component's code as a plugin to try it out, a boolean plugin variable is added to settings (in config.py file) with False as its default value. The run_plugins() function at the "__init__" file of DashAI package launches it with the plugin variable set to True by using an env variable.


DashAI/back/main.py
```diff
@@ -18,6 +19,23 @@
...
+
+# Temporary plugins system endpoint
+@app.get("/update_registry")
+def update_registry():
+    update_plugins(registry_dict_plugins)
+
+
+# Load installed plugins
+if settings.PLUGINS:
+    update_registry()
```
cristian-tamblay, line 37: You shouldnt be calling a FastAPI decorated path operation. Also if you need to startup the plugin system, it shouldnt be on FastAPI main. This file is intended to load endpoints and routes.

DashAI/back/core/config.py
```diff
@@ -32,6 +9,51 @@ class Settings(BaseSettings):
...
+    )
+
+
+# Registries getters
+def get_task_registry():
+    return task_registry
+
+
+def get_model_registry():
+    return model_registry
```
cristian-tamblay, line 59: Maybe move all this code to plugins_system. Its not part of general configuration.

DashAI/back/plugins_system/utils.py
```diff
@@ -0,0 +1,21 @@
...
+    from importlib.metadata import entry_points
+
+
+def update_plugins(registry_dict_plugins):
+    # Retrieve plugins groups (DashAI components)
+    task_plugins = entry_points(group="dashai.plugins.task")
+    model_plugins = entry_points(group="dashai.plugins.model")
+    plugin_groups = {"task": task_plugins, "model": model_plugins}
+
+    for component, plugins in plugin_groups.items():
```
cristian-tamblay, line 15:
```suggestion
    for component, plugins in zip(["task", "model"], [task_plugins, model_plugins]):
```
There is no need to define a plugin_groups dict if you are only iterating on it

plugins/svc_plugin/src/svc_plugin/svc.py
```diff
@@ -0,0 +1,33 @@
...
+# def initialize() -> None:
+#     print("Registrando plugin...")
+#     get_model_registry().register_model(SVC)
+#     return("Plugin registrado!")
+
+def get_class():
+    return SVC
+
+def hello_world():
+    print("Print intermedio...")
```
cristian-tamblay, line 32: If you have a print here linter failed. I think pre-commit is not working on plugins/ folder. Please try to run the linter and remove this print statement.

setup.py
```diff
@@ -1,5 +1,10 @@
 from setuptools import find_packages, setup
 
+# To install extra package from local by using direct references. 
+# In production, this should be installed from pypi
+import os
+path_to_my_project = f"{os.getcwd()}/plugins/tabular_plugin"
```
cristian-tamblay, line 6: This will cause the app to crash if there is no tabular plugin to install. I know this is temporal, but it can give errors, maybe check if directory exists first

DashAI/back/main.py
```diff
@@ -18,6 +19,23 @@
 app.mount(settings.API_V0_STR, api_v0)
 app.mount(settings.API_V1_STR, api_v1)
 
+# Load registries
+task_registry = get_task_registry()
+model_registry = get_model_registry()
+
+registry_dict_plugins = {"model": model_registry, "task": task_registry}
```
cristian-tamblay, line 26: We shouldn't load these here. Also we should have a plugin_registry just like tasks and models.

cristian-tamblay (changes requested):
Most comments refer to the position where the code is called, we are mixing plugin code into FastAPI routes, etc.
We need to keep everything organized for maintenability.

DashAI/back/plugins_system/utils.py
```diff
@@ -0,0 +1,21 @@
+import sys
+
+if sys.version_info < (3, 10):
+    from importlib_metadata import entry_points
+else:
+    from importlib.metadata import entry_points
```
pbadillatorrealba, line 6: Doesn't this do the same thing?

plugins/svc_plugin/src/svc_plugin/svc.py
```diff
@@ -0,0 +1,33 @@
+import json
+#from models.sklearn.sklearn_model import SklearnModel
+from DashAI.back.models.tabular_classification_model import TabularClassificationModel
+from sklearn.svm import SVC as _SVC
+#from base.base_model import model_register
+#from main import model_register
+#from DashAI.back.registries.model_registry import ModelRegistry
+# from DashAI.back.registries.registration import get_model_registry
```
pbadillatorrealba, line 8: To merge I think it is good that you remove all the commented code.
```suggestion
from DashAI.back.models.tabular_classification_model import TabularClassificationModel
from sklearn.svm import SVC as _SVC
```

plugins/svc_plugin/src/svc_plugin/svc.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+class SVC(TabularClassificationModel, _SVC): # Tabular Classification ya hereda de SklearnModel
```
pbadillatorrealba, line 11: Delete comment

```suggestion
class SVC(TabularClassificationModel, _SVC):
```

plugins/svc_plugin/src/svc_plugin/svc.py
```diff
@@ -0,0 +1,33 @@
...
+
+    _compatible_tasks = ["TabularClassificationTask"]
+    MODEL = "SVC"
+    with open(f"DashAI/back/models/parameters/models_schemas/{MODEL}.json") as f:
+        SCHEMA = json.load(f)
+
+# def initialize() -> None:
+#     print("Registrando plugin...")
+#     get_model_registry().register_model(SVC)
+#     return("Plugin registrado!")
```
pbadillatorrealba, line 26: Remove debugging code when you merge.

```suggestion
```

plugins/svc_plugin/src/svc_plugin/svc.py
```diff
@@ -0,0 +1,33 @@
+import json
```
pbadillatorrealba, line 1: Is the svc plugin separated from tabular for testing purposes only?

pbadillatorrealba (commented):
Some code comments.

I would like to mention that at some point we will have to discuss where the plugins will be stored: If in the same repo, as it is in this pr or in different repos.

The advantage of having the base plugins in the same repo is that we will have all the projects (core and plugins) together, but it is a potential source of clutter and complications for the CI/CD.

I take this opportunity to ask, how to publish several packages in pipy from a single repo?

From what I understand, there is monorepo which is an approach to this problem, but I don't know it very well. 

In my opinion, I think it would be best in the long run to have separate repos in the organization.

Some reference: https://www.tweag.io/blog/2023-04-04-python-monorepo-1/

PR #41, opened 2023-04-11
Experiment endpoints

Riul1999:
New Experiment endpoints to manage the experiments of the app.

- Add new route **_/back/experiment_** to the app.
- Add all CRUD methods for endpoint **_/back/experiment_**:
  - GET to get all the available experiment in the DB. Also you can specify the id of the experiment you want to retrieve.
  - POST to create a new experiment and add it to the DB. Dataset_id and task_name are required arguments for the endpoint and no other arguments are accepted.
  - PATCH to modify an existing experiment. Experiment_id is a required argument, dataset_id and task_name are optional argument and no other arguments are acepted.
  - DELETE to remove the experiment from the DB. Experiment_id is a required argument.
- Add new test for the experiment endpoints:
  - Create fixture **_db_** to get access to the testing DB directly.
  - Create **_fixture_dummy_dataset_id_** to create a dummy dataset for the experiment tests.
  - The tests checks all the endpoints and its errors. They also check if the client wants to modify a "read-only" field like step.
- Change **_update_dataset_** endpoint method from PUT to PATCH.
- Add client type to dataset tests.


tests/back/conftest.py
```diff
@@ -11,20 +11,25 @@
...
+    Base.metadata.create_all(bind=engine)
+    return TestingSessionLocal
+
 @pytest.fixture(scope="session", autouse=True)
-def setup_and_delete_db():
+def setup_and_delete_db(session: sessionmaker):
     try:
         shutil.rmtree(f"{USER_DATASETS_PATH}/test_csv", ignore_errors=True)
         shutil.rmtree(f"{USER_DATASETS_PATH}/test_csv2", ignore_errors=True)
         os.remove(TEST_DB_PATH)
```
cristian-tamblay: I think this line is removing the test database after you create it, maybe thats the source of the errors. I removed this line and it works on my local repo
```suggestion
        
```

Riul1999: Yes that was the problem, thanks

cristian-tamblay (approved):

PR #42, opened 2023-04-11
Added production and development envs and json-server

cristian-tamblay:
# Changes
- Added json-server and examples to check database without running backend
- Added multiple environments. yarn start uses .env.development and yarn build uses .env.production
- Fixed some api path errors


DashAI/front/src/utils/index.js
```diff
@@ -4,7 +4,7 @@ export const formatDate = (date) => {
   }
 
   const formattedDate =
-    date.getDate() + "/" + (date.getMonth() + 1) + "/" + date.getFullYear();
-
+    // date.getDate() + "/" + (date.getMonth() + 1) + "/" + date.getFullYear();
+    date;
```
cristian-tamblay: This is giving me error.

DashAI/front/src/components/ExperimentsTable.jsx
```diff
@@ -128,7 +128,7 @@ function ExperimentsTable({ handleNewExperiment }) {
             key="delete-button"
             icon={<DeleteIcon />}
             label="Delete"
-            onClick={handleDeleteExperiment(params.id)}
+            // TODO: onClick={handleDeleteExperiment(params.id)}
```
cristian-tamblay: This is destroying the database. On click handlers receive a function, not an evaluated function

DashAI/front/db.json
```diff
@@ -0,0 +1,191 @@
...
+    },
+    {
+      "id": 3,
+      "dataset_id": 2,
+      "task_name": "Task B",
+      "step": "COMPLETED",
+      "created": "2023-04-11T14:34:00",
+      "last_modified": "2023-04-11T14:34:00"
+    }
+  ],
```
maximi4321:
```suggestion
  "experiment": [
    {
      "id": 1,
      "name": "Experiment A",
      "dataset_id": 1,
      "task_name": "Task A",
      "step": "MODEL_CONFIGURATION",
      "created": "2023-04-11T14:32:00",
      "last_modified": "2023-04-11T14:32:00"
    },
    {
      "id": 2,
      "name": "Experiment B",
      "dataset_id": 1,
      "task_name": "Task A",
      "step": "EXECUTION",
      "created": "2023-04-11T14:33:00",
      "last_modified": "2023-04-11T14:33:00"
    },
    {
      "id": 3,
      "name": "Experiment C",
      "dataset_id": 2,
      "task_name": "Task B",
      "step": "COMPLETED",
      "created": "2023-04-11T14:34:00",
      "last_modified": "2023-04-11T14:34:00"
    }
  ],
```
I think the names for each experiment are missing.

DashAI/front/package.json
```diff
@@ -31,7 +34,8 @@
     "start": "craco start",
     "build": "craco build",
     "test": "craco test",
-    "eject": "react-scripts eject"
+    "eject": "react-scripts eject",
+    "json-server": "json-server db.json --routes routes.json --port 3001"
```
maximi4321, line 36: I haven't researched much on this, so please ignore this if it introduces too many complications. But I think it could be useful to have the option to use `.env.production` when starting with `yarn start`. 

Maybe adding something like this:
`"start:custom": "NODE_ENV=production craco start" `
 so that when doing `yarn start:custom`, `.env.production` can be used.

maximi4321:
In general, it's working well for me, but I have some small comments.

- I think the `name` attribute for each experiment is missing, and some of the names in the JSON file do not match those used in the experiments table, which makes the experiments table almost empty.
- I couldn't get it to work in production with `yarn build`. For example, when I try to create a new experiment, it doesn't receive the data from the available tasks and their descriptions.
- Just as an idea, as I haven't researched much on the topic. It might be good to be able to use the values from `.env.production` when using `yarn start` to be able to test some functionality with the real database.

cristian-tamblay:
> * Just as an idea, as I haven't researched much on the topic. It might be good to be able to use the values from `.env.production` when using `yarn start` to be able to test some functionality with the real database.

When using yarn start, .env.development is loaded. We could maybe set json-server port to 8000 so its the same as the backend one

cristian-tamblay:
> * I think the `name` attribute for each experiment is missing, and some of the names in the JSON file do not match those used in the experiments table, which makes the experiments table almost empty.

Added name to db table experiment, I'll need to fix experiment PR files.

> * I couldn't get it to work in production with `yarn build`. For example, when I try to create a new experiment, it doesn't receive the data from the available tasks and their descriptions.

This didnt work for me neither in yarn build or yarn start. The endpoint is not working yet

> * Just as an idea, as I haven't researched much on the topic. It might be good to be able to use the values from `.env.production` when using `yarn start` to be able to test some functionality with the real database.

I tested some solutions to this, `yarn start --env=production` but i doesn't work. I think the best way to test production settings in development is just to copy the .env.production into .env.development. This shouldn't be a common thing.

maximi4321 (approved):

README.rst
```diff
@@ -100,6 +100,11 @@ Then, launch the front-end development server by running the following command:
 
     $ yarn start
 
+If you want to use a test database and not the backend, you can run on another terminal
```
pbadillatorrealba: I would change this sentence to something more explicit.

```suggestion
If you want to launch the front-end test server (without launching the backend) with dummy data, execute:
```

DashAI/front/db.json
```diff
@@ -0,0 +1,194 @@
+{
```
pbadillatorrealba: Wouldn't it be good to call this file dummyData or testData or testDB ?

pbadillatorrealba (changes requested):
I think this will be a very good contribution! I left small comments in the readme.

Question: what is the utility of declaring all paths in .env files? Is it strictly necessary? 

I think a lot of things are duplicated there, which could eventually lead to errors.

README.rst
cristian-tamblay: :ok_hand:

DashAI/front/db.json
cristian-tamblay: :ok_hand:

cristian-tamblay:
> Question: what is the utility of declaring all paths in .env files? Is it strictly necessary?
> 
> I think a lot of things are duplicated there, which could eventually lead to errors.

Yes, there are a lot of duplicated routes that will be removed in the future when the endpoints are completly moved to src/api in typescript. Its kinda a legacy compatibility meanwhile.

Both of these files will only require 1 line setting REACT_API_URL to use it here:

```typescript
const api: AxiosInstance = axios.create({
  baseURL: process.env.REACT_APP_API_URL,
});
export default api;
```

pbadillatorrealba (approved):

PR #43, opened 2023-04-12
Task Endpoint

Riul1999:
New endpoint to get all the registered tasks and its descriptions.

- Add new route **_/back/task** to the app.
- Add all CRUD methods for endpoint **_/back/task**:
  - GET to get all the available tasks.
  - POST, PATCH and DELETE methods where not implemented and they raise an 501 error when called.
- Add new test for the task endpoints:
  - The test check the endpoint send to the user all the task available in the **_register_task_** register.


DashAI/back/api/api_v1/api.py
```diff
@@ -1,6 +1,7 @@
 from fastapi import APIRouter
 
-from DashAI.back.api.api_v1.endpoints import datasets
+from DashAI.back.api.api_v1.endpoints import datasets, tasks
 
 api_router_v1 = APIRouter()
 api_router_v1.include_router(datasets.router, prefix="/dataset")
+api_router_v1.include_router(tasks.router, prefix="/task")
```
cristian-tamblay, line 8: There is a conflict here with recent push on staging branch, please fix it (merge staging and push the fix)

DashAI/back/tasks/tasks_schemas/BaseTask.json
```diff
@@ -0,0 +1,7 @@
+{
+	"class": "BaseTask",
+	"name": "Base",
+	"help": "Base class for tasks.",
+	"description": "Base class for tasks.",
+	"type": "object"
+}
```
cristian-tamblay: I think BaseTask is not something that should have a schema

DashAI/back/tasks/tasks_schemas/TabularClassificationTask.json
```diff
@@ -1,19 +1,7 @@
...
-			"images": ["/info_images/csv_files.png"]
-		},
-		{
-			"class": "JSONDataLoader",
-			"name": "JSON Data",
-			"help": "Use JSON files to upload the data. You can use a .json file or multiple .json files in a .zip file.",
-			"description": "You can upload your data in a .json file or in multiple .json files in a .zip, where you can have the definition of splits in folders as shown above. \n If you only have one file or multiple files without the folder definition for the splits, you can set them up later. \n The JSON files have to follow the format that specify in the last figure, the data have to be in a key, that contains an array with the samples of the data or rows in JSON objects, where each one have the features of data or columns like key/value pairs. \n Make sure that all the JSON files have the same features and format.",
-			"images": ["/info_images/json_files.png", "/info_images/json_format.svg"]
-		}
-	],
```
cristian-tamblay, line 17: I like the idea of removing these from here, but we need to save those on another location. They are still useful.

cristian-tamblay (changes requested):

DashAI/back/tasks/tasks_schemas/TabularClassificationTask.json
Riul1999: I will add it to the dataloaders endpoint PR

DashAI/back/api/api_v1/endpoints/tasks.py
```diff
@@ -0,0 +1,55 @@
...
+
+    register_tasks = task_registry.registry
+    if not register_tasks:
+        raise HTTPException(
+            status_code=status.HTTP_404_NOT_FOUND, detail="No tasks found"
+        )
+    all_tasks = defaultdict(dict)
+    for task_name, task_cls in register_tasks.items():
+        all_tasks[task_name] = task_cls.get_schema()
+    return all_tasks
```
pbadillatorrealba: It is not necessary in this case to use defaultdict since you know the name of all the keys beforehand (and you are not using any object factory (tasks in this case) by default).

I propose this change using dict comprehensions:

```suggestion
    all_tasks = {
        task_name: task_cls.get_schema()
        for task_name, task_cls in register_tasks.items()
    }
    return all_tasks
```

DashAI/back/tasks/base_task.py
```diff
@@ -1,3 +1,4 @@
+import json
```
pbadillatorrealba, line 1: Adds a logger to record when a schema is not found (see comment below).

```suggestion
import json
import logging

# all other imports

logger = logging.getLogger(__name__)

```

DashAI/back/tasks/base_task.py
```diff
@@ -59,6 +60,13 @@ def add_compatible_component(
 
         cls.compatible_components[registry_for.__name__][component.__name__] = component
 
+    @classmethod
+    def get_schema(self) -> dict:
+        """
+        This method load the schema JSON file asocciated to the task.
+        """
+        return json.load(open(f"DashAI/back/tasks/tasks_schemas/{self.name}.json", "r"))
```
pbadillatorrealba: Here you have to use a context manager to close the file at the end of the read operation.
Also, add a try/except to catch the case where the schema file is not found.

```suggestion
        try:
            with open(f"DashAI/back/tasks/tasks_schemas/{self.name}.json", "r") as f:
                schema = json.load(f)
            return schema
        except FileNotFoundError:
            logger.exception(
                (
                    f"Could not load the schema for {self.__name__} : File DashAI/back"
                    f"/tasks/tasks_schemas/{self.name}.json not found."
                )
            )
            return {}
```

tests/back/api/test_tasks.py
```diff
@@ -0,0 +1,13 @@
...
+
+def test_get_all_tasks(client: TestClient):
+    task_names = task_registry.registry.keys()
+    response = client.get("/api/v1/task/")
+    assert response.status_code == 200, response.text
+    data = response.json()
+    for task_name, task_data in data.items():
+        assert task_name in task_names
+        assert task_data["class"] in task_names
+        assert task_name == task_data["class"]
```
pbadillatorrealba: I am concerned here that the test is not very deterministic as the tasks you load by default in the app may vary (and if all goes well, eventually not load any in the core).

Here I would suggest using a task registry from scratch and adding test tasks by hand:

```suggestion
from DashAI.back.registries.task_registry import TaskRegistry
from DashAI.back.tasks.base_task import BaseTask
from fastapi.testclient import TestClient
from DashAI.back.core import config


class TestTask1(BaseTask):
    ...


class TestTask2(BaseTask):
    ...


test_task_registry = TaskRegistry(initial_components=[TestTask1, TestTask2])
# replace the default tasks with the previously test tasks
config.task_registry._registry = test_task_registry ._registry


def test_get_all_tasks(client: TestClient):
    task_names = task_registry.registry.keys()
    response = client.get("/api/v1/task/")

    assert response.status_code == 200, response.text

    data = response.json()
    assert len(data.keys()) == 2
    for task_name, task_data in data.items():
        assert task_name in task_names
        assert isinstance(task_data, dict)
```

Note that these changes only work if the above corrections are accepted.

pbadillatorrealba (changes requested):
I left some changes regarding syntax, error handling and a relatively big one regarding how registries and related apis are tested. Hopefully we can discuss that point :)

pbadillatorrealba (approved):

cristian-tamblay (approved):

PR #44, opened 2023-04-19
Dataloaders endpoint

Riul1999:
New endpoint to get all the registered dataloader linked with a task and its descriptions.

- Add new route **_/back/dataloader** to the app.
- Add all CRUD methods for endpoint **_/back/dataloader**:
  - GET to get all the available dataloaders linked with the given task.
  - POST, PATCH and DELETE methods where not implemented and they raise an 501 error when called.
- Add new test for the dataloader endpoints:
  - The test check the endpoint send to the user all the dataloaders available in the **_register_dataloader_** register linked with a test Task.
  - Create DataloaderRegsiter to save all the available Dataloaders and its linked Tasks.
  - Init DataloaderRegister with CSVDataloader and JSONDataloader in the config file.
  - Add description schemas to dataloaders to save all their information.
  - Add method get_schema to BaseDataloader to get the description schema of the dataloader.
  - Add variable _compatible_tasks to TabularDataloader in order to link them with TabularClassificationTask.


DashAI/back/registries/dataloader_registry.py
```diff
@@ -0,0 +1,9 @@
+from DashAI.back.dataloaders import BaseDataLoader
+from DashAI.back.registries.base_registry import BaseRegistry
+from DashAI.back.registries.register_mixins import TaskComponentMappingMixin
+
+
+class DataloaderRegistry(BaseRegistry, TaskComponentMappingMixin):
+    """Centralized registry for models."""
```
pbadillatorrealba:
```suggestion
    """Centralized registry for dataloaders."""
```

tests/back/api/test_dataloaders.py
```diff
@@ -0,0 +1,69 @@
+import pytest
+
+from DashAI.back.registries import DataloaderRegistry, TaskRegistry
+from DashAI.back.tasks import BaseTask
+from DashAI.back.dataloaders import BaseDataLoader
+from fastapi.testclient import TestClient
+from DashAI.back.core import config
+
+default_dataloader_dict = {'foo': 'bar'}
```
pbadillatorrealba: What is the purpose of this dict?

tests/back/api/test_dataloaders.py
```diff
@@ -0,0 +1,69 @@
...
+    test_task_registry = TaskRegistry(initial_components=[TestTask])
+
+    test_dataloader_registry = DataloaderRegistry(
+        initial_components=[TestDataloader1, TestDataloader2],
+        task_registry=test_task_registry,
+    )
+    # replace the default dataloaders with the previously test dataloaders
+    config.dataloader_registry._registry = test_dataloader_registry._registry
+    config.dataloader_registry.task_component_mapping = test_dataloader_registry.task_component_mapping
+    yield test_dataloader_registry
```
pbadillatorrealba, line 53: What is the purpose of this `yield`? Is there a generator??

pbadillatorrealba (commented):
Some comments and questions.

tests/back/api/test_dataloaders.py
cristian-tamblay: I think this yield separates the setup from the teardown on this module

DashAI/back/api/api_v1/endpoints/dataloaders.py
```diff
@@ -0,0 +1,54 @@
...
+    -------
+    List[JSON]
+        List of dataloaders JSONs
+    """
+
+    register_dataloaders = dataloader_registry.task_to_components(task_name)
+    if not register_dataloaders:
+        raise HTTPException(
+            status_code=status.HTTP_404_NOT_FOUND, detail="No dataloaders found"
+        )
```
cristian-tamblay: We agreed with the frontend to return an empty list `[]` when there is no object found in the database. Maybe in this case return an empty json `{}`.

pbadillatorrealba: I don't know if this is a good idea since the endpoint always expects a list to be returned, even if it is empty. Returning a json may cause an exception to be raised on the front end, since it is not the type of data it was expecting.

cristian-tamblay: Fixed, now it returns [] or a list of objects

DashAI/back/registries/dataloader_registry.py
cristian-tamblay: :ok_hand:

tests/back/api/test_dataloaders.py
cristian-tamblay: This dict was meant to be a test for a dict like this:
```
{
    "class": "JSONDataLoader",
    "name": "JSON Data",
    "help": "Use JSON files to upload the data. You can use a .json file or multiple .json files in a .zip file.",
    "description": "You can upload your data in a .json file or in multiple .json files in a .zip, where you can have the definition of splits in folders as shown above. \n If you only have one file or multiple files without the folder definition for the splits, you can set them up later. \n The JSON files have to follow the format that specify in the last figure, the data have to be in a key, that contains an array with the samples of the data or rows in JSON objects, where each one have the features of data or columns like key/value pairs. \n Make sure that all the JSON files have the same features and format.",
    "images": ["/info_images/json_files.png", "/info_images/json_format.svg"],
	"type": "object"
}
```
I replaced it with a more explicit `{'class':'DataLoader1'}` inside the DataLoader, get_schema

pbadillatorrealba (approved):

PR #45, opened 2023-04-28
Parameter form refactor

maximi4321:
# Summary
Refactor of parameter form: the code is organized into separate components, the use of Bootstrap in components is replaced with Material-UI, and rendering a `ParameterForm` is simplified.

# Details
- The `ConfigurableObject` folder is created under `components`, where most of the modifications of this PR are located.
- The `FormInputs.jsx` component is created, containing all the inputs that can be rendered in the parameter form.
- Two components are created to simplify the render: 
  - `MainForm.jsx` replaces the previous version of `ParameterForm`
  -  `ParameterForm.jsx` is now a wrapper for `MainForm.jsx` that stores the necessary states (e.g., default values) so that now only a JSON needs to be provided to render a form.
- The code that handles the dataloaders is organized into components found in the `ConfigurableObject` folder (`ParamsModal.jsx` and `SplitsParams.jsx`).
- Both the form inputs and modals are changed to Material-UI.

# Issues
- There are still some warnings associated with the default values of the form inputs.

# TODO
- Refactor the logic to validate the form, as there is currently no validation.


DashAI/front/src/components/ConfigurableObject/FormInputs.jsx
```diff
@@ -0,0 +1,377 @@
...
+  onChange: PropTypes.func.isRequired,
+  description: PropTypes.string.isRequired,
+  error: PropTypes.string,
+};
+Integer.defaultProps = {
+  error: undefined,
+};
+
+// This field is designed to handle non-integer numeric inputs
+export function Number({ name, value, onChange, description, error }) {
```
pbadillatorrealba: Move to `front/src/components/ConfigurableObject/Inputs/NumberInput.jsx`

```suggestion
export function NumberInput({ name, value, onChange, description, error }) {
```

DashAI/front/src/components/ConfigurableObject/FormInputs.jsx
```diff
@@ -0,0 +1,377 @@
...
+import PropTypes from "prop-types";
+import { getDefaultValues } from "../../utils/values";
+import Subform from "./Subform";
+import { styled } from "@mui/material/styles";
+
+const Input = styled(TextField)(({ theme }) => ({
+  minWidth: "20vw",
+}));
+
+export function Integer({ name, value, onChange, description, error }) {
```
pbadillatorrealba: Move to `front/src/components/ConfigurableObject/Inputs/IntegerInput.jsx`

```suggestion
export function IntegerInput({ name, value, onChange, description, error }) {
```

DashAI/front/src/components/ConfigurableObject/FormInputs.jsx
```diff
@@ -0,0 +1,377 @@
...
+  value: PropTypes.number.isRequired,
+  onChange: PropTypes.func.isRequired,
+  description: PropTypes.string.isRequired,
+  error: PropTypes.string,
+};
+Number.defaultProps = {
+  error: undefined,
+};
+
+export function Float({ name, value, onChange, description, error }) {
```
pbadillatorrealba: Move to `front/src/components/ConfigurableObject/Inputs/FloatInput.jsx`

```suggestion
export function FloatInput({ name, value, onChange, description, error }) {
```

DashAI/front/src/components/ConfigurableObject/FormInputs.jsx
```diff
@@ -0,0 +1,377 @@
...
+  value: PropTypes.number.isRequired,
+  onChange: PropTypes.func.isRequired,
+  description: PropTypes.string.isRequired,
+  error: PropTypes.string,
+};
+Float.defaultProps = {
+  error: undefined,
+};
+
+export function Text({ name, value, onChange, error, description }) {
```
pbadillatorrealba: Move to `front/src/components/ConfigurableObject/Inputs/TextInput.jsx`

```suggestion
export function TextInput({ name, value, onChange, error, description }) {
```

DashAI/front/src/components/ConfigurableObject/FormInputs.jsx
```diff
@@ -0,0 +1,377 @@
...
+  onChange: PropTypes.func.isRequired,
+  description: PropTypes.string.isRequired,
+  error: PropTypes.string,
+};
+Text.defaultProps = {
+  value: "",
+  error: undefined,
+};
+
+export function Select({ name, value, onChange, error, description, options }) {
```
pbadillatorrealba: Move to `front/src/components/ConfigurableObject/Inputs/SelectInput.jsx`

```suggestion
export function SelectInput({ name, value, onChange, error, description, options }) {
```

DashAI/front/src/components/ConfigurableObject/FormInputs.jsx
```diff
@@ -0,0 +1,377 @@
...
+  onChange: PropTypes.func.isRequired,
+  description: PropTypes.string.isRequired,
+  error: PropTypes.string,
+  options: PropTypes.arrayOf(PropTypes.string).isRequired,
+};
+Select.defaultProps = {
+  error: undefined,
+};
+
+export function Boolean({ name, value, onChange, error, description }) {
```
pbadillatorrealba: Move to `front/src/components/ConfigurableObject/Inputs/BooleanInput.jsx`

```suggestion
export function BooleanInput({ name, value, onChange, error, description }) {
```

Also, it is necessary to use a FormControl parent component?

DashAI/front/src/components/ConfigurableObject/FormInputs.jsx
```diff
@@ -0,0 +1,377 @@
...
+  value: PropTypes.bool.isRequired,
+  onChange: PropTypes.func.isRequired,
+  description: PropTypes.string.isRequired,
+  error: PropTypes.string,
+};
+Boolean.defaultProps = {
+  error: undefined,
+};
+
+function ClassInput({
```
pbadillatorrealba: Move to `front/src/components/ConfigurableObject/Inputs/ClassInput.jsx`

DashAI/front/src/components/ConfigurableObject/FormInputs.jsx
```diff
@@ -0,0 +1,377 @@
...
+      return <Text {...commonProps} />;
+    case "boolean":
+      return <Boolean {...commonProps} />;
+    case "float":
+      return <Float {...commonProps} />;
+    default:
+      return (
+        <Typography variant="p" sx={{ color: "red" }}>
+          {`Not a valid parameter type ${type}`}
+        </Typography>
```
pbadillatorrealba: I think the correct thing to do in this case is for the code to throw an exception.

```suggestion
      throw new Error(`Error while rendering ${objName}: ${type} is not a valid parameter type.`)
```

DashAI/front/src/components/ConfigurableObject/FormInputs.jsx
```diff
@@ -0,0 +1,377 @@
...
+      PropTypes.object,
+    ])
+  ),
+};
+
+ClassInput.defaultProps = {
+  formDefaultValues: { emptyDefaultValues: true },
+};
+
+export function genInput(objName, paramJsonSchema, formik, defaultValues) {
```
pbadillatorrealba: I would give this class a more descriptive name, such as `ConfigurableObjectRenderer` or `ConfigurableObjectFormRenderer`.

```suggestion
export function ConfigurableObjectRenderer(objName, paramJsonSchema, formik, defaultValues) {
```

pbadillatorrealba: I would give this class a more descriptive name, such as `FormRenderer`, `ConfigurableObjectRenderer` or `ConfigurableObjectFormRenderer`.

```suggestion
export function FormRenderer(objName, paramJsonSchema, formik, defaultValues) {
```

If this change is made, the name of the module would also have to be changed.

DashAI/front/src/components/ConfigurableObject/MainForm.jsx
```diff
@@ -0,0 +1,73 @@
...
+      {submitButton && (
+        <Button
+          style={{ float: "right" }}
+          size="large"
+          onClick={formik.handleSubmit}
+        >
+          Save
+        </Button>
+      )}
+    </div>
```
pbadillatorrealba: Unnecessary `div`

```suggestion
      {genInput("", parameterSchema, formik, defaultValues)}
      {extraOptions}
      {submitButton && (
        <Button
          style={{ float: "right" }}
          size="large"
          onClick={formik.handleSubmit}
        >
          Save
        </Button>
      )}
```

pbadillatorrealba (changes requested):
The PR is too long to review in detail!

I would strongly recommend splitting it (at least) in two: 
1. The changes to the forms.
2. The changes to datasets.

The FormInputs.jsx file has many functions that render components and a lot of functionality. At first glance, it is very complex and I had a hard time understanding what does what.
The correct thing to do would be to separate the classes that make forms into distinct modules. 
I left a comment on each of the functions that I think should be moved.


Some other questions:

- What is the flow of use of the forms? In other words, what function should I instantiate to get only one form? 
- Why split has a special case? Does it not meet the specifications?
- Why `ParameterForm` has a call to the backend?
- Why every form component must hace a div parent? if it is not necessary, then we should delete them.
- What is the purpose of Number, if there are inputs for Float and for Integer?

maximi4321:
> The PR is too long to review in detail!
> 
> I would strongly recommend splitting it (at least) in two:
> 
> 1. The changes to the forms.
> 2. The changes to datasets.
> 
Sorry! there was an error. The dataset changes (2.) were part of another PR that has already been merged into staging. That has been resolved, and now only the parameter form commits should remain (1.)

> The FormInputs.jsx file has many functions that render components and a lot of functionality. At first glance, it is very complex and I had a hard time understanding what does what. The correct thing to do would be to separate the classes that make forms into distinct modules. I left a comment on each of the functions that I think should be moved.
> 

The latest commit incorporates the suggestions: the code has been organized and names have been changed to be more descriptive.

> Some other questions:
> 
> * What is the flow of use of the forms? In other words, what function should I instantiate to get only one form?
>
The only component that should be used to render a parameter form is `ParameterForm.jsx`, which is responsible for calling all the necessary logic to make it work.

The minimum requirement for the component to work is the JSON that comes from the API.
Here is an example: `<ParameterForm parameterSchema={schema} />`

 There are other parameters such as `onFormSubmit`, that is a function `(values) => {}`  that handles what to do when the form is submitted. (values is a json with the values of the parameters, this json is provided by formik)
> * Why split has a special case? Does it not meet the specifications?
>
No, `splits` is a special case that is sent in the JSON of the data loaders in a different way, not as properties but as the `split` property itself. This can be seen in the file: `DashAI/DashAI/back/dataloaders/params_schemas/csvdataloader.json`.

Since there was already logic to support this type of JSON and due to time constraints, it was decided to keep that existing logic.
> * Why `ParameterForm` has a call to the backend?
>
I think you're referring to when the function `getFullDefaultValues` is called. This is because in earlier stages, the functionality was to allow the user to choose a model and run it without configuring it, using its default values.

Since there are parameters that are themselves other forms (recursive parameters), calling the backend is necessary to obtain _all_ the default values.

For example, for `numericalWrapperForText`, one of its parameters is a numerical configuration model. When the JSON of `numericalWrapperForText` is obtained, the backend is called again to obtain the default values of one of the numerical models to complete all the default values that `numericalWrapperForText` needs and to be able to run the model without configuring it.

> * Why every form component must hace a div parent? if it is not necessary, then we should delete them.
>
I tried using a `Fragment` to return both the input and the tooltip, but that made them appear on different "rows" (the tooltip was placed below the input). This problem was fixed by using a `div` to contain both. I don't know if there is another way to solve that problem.
> * What is the purpose of Number, if there are inputs for Float and for Integer?
>
That was also kept from the previous implementation due to time constraints. This is because the JSON of both dataloaders and models have types: integer, float, and number.

To fix it, both the frontend and backend need to be modified

DashAI/front/src/components/ConfigurableObject/MainForm.jsx
pbadillatorrealba, line 1: This component seems to me to be a bit redundant. Why not merge it with ParameterForm.jsx?

DashAI/front/src/components/ConfigurableObject/MainForm.jsx
```diff
@@ -0,0 +1,73 @@
...
+    onSubmit: (values) => {
+      onFormSubmit(values);
+    },
+  });
+  React.useEffect(() => {
+    // get current values of an input
+    if (getValues !== null && typeof getValues !== "undefined") {
+      getValues[1](formik.values[getValues[0]]);
+    }
+  }, [formik.values]);
```
pbadillatorrealba: What does this code do?
This is very difficult to understand :(
Could you add a little description as a comment on the function signature?

DashAI/front/src/components/ConfigurableObject/ParameterForm.jsx
```diff
@@ -0,0 +1,65 @@
...
+  // TODO: find a better way because this fails if the new parameterSchema has a parameter with the same name as the previous parameterSchema.
+  const checkDefaultValues = () => {
+    if (defaultValues === null) {
+      return false;
+    }
+    return Object.keys(parameterSchema.properties)[0] in defaultValues;
+  };
+  async function getFormInitialValues() {
+    setLoading(true);
+    const values = await getFullDefaultValues(parameterSchema);
```
pbadillatorrealba: This API call should be in a try-catch-finally. In case it fails, it should notify to the user through a snackbar or something similar.

DashAI/front/src/utils/values.jsx
pbadillatorrealba, line 1: Could you post a couple of descriptions of what the functions in this file do?
It's hard to understand the code, especially the ifs and the cases that each one handles.

Also, I take this opportunity to ask, knowing that this code makes calls to the backend, could you move this code to the APIs?

pbadillatorrealba (changes requested):
I left other comments regarding the code.

I still have doubts as to why there are `MainForm` and `SubForm`. Can this be simplified?

I don't know if you could explain the rendering process of these components with a flowchart?

Also, I think it is important to leave very detailed for the rest of the devs what ParameterForm does and how it works. Could you add documentation to this and its parameters (using JSDoc - in vscode it generates you a template by typing //**)?

Other topics:
- Regarding the code style, to maintain readability I think it is a good idea to separate functions with spaces. Could you separate functions and others by spaces?
- Could you comment briefly what do the implemented useEffect do?

DashAI/front/src/components/ConfigurableObject/MainForm.jsx
maximi4321: Yes, the components `MainForm.jsx` and `Subform.jsx` are both used to render a form.

Both components use the hooks `useFormik` and `useEffect`, but with different functionalities. This is why i initially preferred to separate them into two components. The main differences in the components are:

- `Subform.jsx` adds a `choice` property to its default values  in the `useFormik` hook to indicate to the backend the option the user selected for a recursive parameter.
- the submit of the values handled by the `useFormik` hook is different.
  - `MainForm.jsx` uses the `onSubmit`  property to retrieve the values from the entire form and to perform an action with those values like sending them to the backend.
  - `SubForm.jsx` updates its values through a `useEffect` that uses the `setFieldValue` function.
- `MainForm.jsx` includes an `extraOptions` property to handle additional functionality, specifically for dataloaders that have different JSON properties like `splits`.

In summary: `MainForm.jsx` handles the submit of the entire form and also manages extra options (dataloaders) , while `Subform.jsx` only updates the values of the parameters.

Maybe the components can be merged  using if-else conditions inside the `useFormik` and `useEffect` hooks.

DashAI/front/src/components/ConfigurableObject/MainForm.jsx
maximi4321: Sorry, I did not make any changes to that part of the code  due to time constraints.
That code is used mainly for the dataloaders, to manage when the form has `extraOptions`.

getValues is an array: `[name_of_parameter, function]`
This is used to get the value of a specific parameter and trigger additional behavior in the form.

One concrete example of this can be seen in the dataloader forms, specifically with the `splits_in_folder` parameter. If the value of `splits_in_folder` is set to false (as shown in the left form of the image), the user will be required to specify the splits, and additional fields will be displayed. However, if the value of splits_in_folder is set to true (as shown in the right form of the image), these additional fields disappear as they are no longer needed.

splits_false

DashAI/front/src/utils/values.jsx
maximi4321: > Also, I take this opportunity to ask, knowing that this code makes calls to the backend, could you move this code to the APIs?

I moved the function that makes backend calls to `/api`, but I had to define some additional types and modify the function a bit (added an if-else). These changes will go in another PR that depends on this one.

cristian-tamblay (approved):

pbadillatorrealba (approved):

PR #46, opened 2023-04-28
Remove .env dependencies

maximi4321:
The dependency on the `.env` files is removed, and the modifications to the code are minimal to replicate exactly what was there before.

- The `oldEndpoints.ts` file is added to handle the old endpoints (e.g. the endpoint to get schemas for the parameter form). Some of the calls are not tested because it is not possible to do so in the current state of the application.
- The `datasets.ts` file is added for calls corresponding to datasets, which are mainly made for the functionalities of the datasets table.
- All environment variables are removed except `REACT_APP_API_URL`.
- Components with unused code that also had `.env` dependencies (`ExperimentPipeline.jsx` and `MainForm.jsx`) are removed.


DashAI/front/src/App.jsx
pbadillatorrealba, line 1: This navigation modification should go in another pr, as they are different concepts.

pbadillatorrealba (changes requested):

pbadillatorrealba:
This pr was divided into #52 , #53 and #54.
It will be closed.

PR #47, opened 2023-05-02
Update app navegation

pbadillatorrealba:
This pr: 

- Remove the error page in case the API is not found (now every call should notify about this problem).
- Remove the Grid that encapsulates the pages. This is because the styles of each page should be defined in its respective file and not in the global path container.


cristian-tamblay (approved):
I love 1 file PRs

PR #51, opened 2023-05-02
Update frontend to use new task endpoint

maximi4321:
Use the new endpoint for tasks in the frontend.
- Define `getTasks()` function in `tasks.ts` to retrieve tasks using the new endpoint.
- Add `tasks.ts` to the types folder to define the properties expected in the tasks data.
- Modify `SchemaList.jsx` to use `getTasks()` function for retrieving tasks.
- Add conditional statement to change between new task endpoint and legacy endpoint, depending on the context (tasks need the new endpoint, dataloaders need the legacy endpoint)

### DIFF
--- a/DashAI/front/src/api/task.ts
+++ b/DashAI/front/src/api/task.ts
@@ -1,0 +1,7 @@
+import api from "./api";
+import type { ITask } from "../../types/task";
+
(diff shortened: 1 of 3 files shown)


cristian-tamblay (approved):
Working as intended. Next step is using dataloader endpoint

PR #62, opened 2023-05-10
Back Registry endpoint

Riul1999:
New endpoint to get all the registered objects and its descriptions.

- Add new route **_/back/registry/relationship/{source_type}/{task_name}** and **_/back/registry/inheritance/{source_type}/parent_name?** to the app.
-  **_/back/registry/relationship/{source_type}/{task_name}** give all the objects of type _source_type_ related to the Task with name _task_name_.
- **_/back/registry/inheritance/{source_type}/parent_name?** give all the objects of type _source_type_ that inherits from the object with name _parent_name_. If no parent_name is given, the endpoints gives all the objects with the type _source_type_.
- Merge and refactor test_dataloaders and test_tasks into test_register.
- Create dictionary name_registry_mapping to map type with its registry.
- Add method _parent_to_components_ to the BaseRegistry class to get all the components that inherits from an specific parent component.


pbadillatorrealba (dismissed):
I think it would be better to use [Query parameters](https://fastapi.tiangolo.com/tutorial/query-params/) instead of Path parameters for the new endpoints.

The idea behind this is to think that the Registry is a resource that acts similar to a database, rather than like a file directory (which is how it is implemented now using path parameters). In this case, the only path needed would be the one indicating access to the registry.
On the other hand, queries allow you to filter and sort the information, which would allow you to select only the relevant components from the registry (Source: [When do I use path params vs. query params in a RESTful API?](https://www.alibabacloud.com/topic-center/dev-faq/ghgfdxwc43-when-do-i-use-path-params-vs-query-params-in-a-restful-api)).

Here is how I think these endpoints should be implemented using query parameters.

- You can get every component registered in the registry using a get over `/registry`

- You can get the schema of a particular class through a get to `/registry/{id}` with a parameter `id`. For example `/registry/KNeighborsClassifier`. In this case the ids are supposed to be unique.

- You can get the schema of all the components associated to a certain task through a `task_name` parameter. Example: `/registry?task_name=TabularClassification`.

- You can get a component (remember we agreed to use that name) of a specific type using the `component_type` parameter. Example `/registry?component_type=model` (here we could make a mechanism to omit the `"Base"`).

Then, you can combine these parameters (without a mandatory order) according to your needs:

For example, to ask for all the models associated to the task classification you can directly use `/registry?task_name=TabularClassification&component_type=model`.
The above would replace `get_related_objects`.

On the other hand, to get all models, you could simply use `/registry?component_type=model`. This would replace `get_child_objects`.

The advantage of this type of endpoint is that all the logic is in a single generalizable function.
Then, the enpoint implementation boils down to obtaining only the relevant objects from the Registries using the queries.

pbadillatorrealba:
The last commit implements the new design of the API (called now components, which is the noun name of the API objects, instead of registry).

The major changes are: 

- Implement `get_components, method` that retrieves components from the registry with query options.
- Implement a `get_component_by_id` that returns a specific component using its id.
- Reimplemented parent to component function based on base classes names.
- Implemented several tests for the previous described functionality.

@Riul1999 , I can't set you as reviewer but could you review the changes?

DashAI/back/api/api_v1/endpoints/components.py
```diff
@@ -0,0 +1,157 @@
...
+) -> list[dict]:
+    """Retrieves components from the register.
+
+    The components returned will depend on the parameters of the request.
+    If all parameters are None, then the method will return all registered components.
+
+    Parameters
+    ----------
+    component_type : str | None, optional
+        If specified, the function will return only the components related with the
```
inunezn7:
```suggestion
        If specified, the function will return only the components belonging to that
```

DashAI/back/api/api_v1/endpoints/components.py
```diff
@@ -0,0 +1,157 @@
...
+        component type (e.g., task, model, dataloader, etc...), by default None.
+    task_name : str | None, optional
+        If specified, the function will return only the components related with
+        the task (e.g., TabularClassification, Translation), by default None.
+    component_parent : str | None, optional
+        If specified, the function will return only the components that inheirts the
```
inunezn7:
```suggestion
        If specified, the function will return only the components that inherit from the
```

DashAI/back/api/api_v1/endpoints/components.py
```diff
@@ -0,0 +1,157 @@
...
+        task_related_components = []
+        for component_type in name_registry_mapping:
+            if component_type != "task":
+                for component_name in name_registry_mapping[
+                    component_type
+                ].task_to_components(task_name):
+                    task_related_components.append(component_name)
+
+        # filter
+        selected_components = selected_components = {
```
inunezn7:
```suggestion
        selected_components = {
```

DashAI/back/api/api_v1/endpoints/components.py
```diff
@@ -0,0 +1,157 @@
...
+
+@router.post("/", status_code=status.HTTP_201_CREATED)
+async def upload_component():
+    raise HTTPException(
+        status_code=status.HTTP_501_NOT_IMPLEMENTED, detail="Method not implemented"
+    )
+
+
+@router.delete("/")
+async def defete_component():
```
inunezn7:
```suggestion
async def delete_component():
```

inunezn7 (approved):
I just pointed out a few typos.
One thing I was thinking about is if we will need that the get_components_by_parent could retrieve the children and also the grandchildren, grand-grandchildren, etc, in case they are concrete components. Nevertheless, this could be discussed later and resolved if necessary based on the changes in this PR.

pbadillatorrealba:
This pr will be reactivated when the new register implementation is ready.

PR #64, opened 2023-05-12
Back/run endpoint

Riul1999:
New Run endpoints to manage the run of the app.

- Add new route **_/back/run** to the app.
- Add all CRUD methods for endpoint **_/back/run**:
  - GET to get an specific run from the DB. Also you can specify an experiment id to get all the runs associated with that experiment (the endpoint for that is **_/back/run/list/{experiment_it}**)
  - POST to create a new run and add it to the DB. Experiment_id, model_name, run_name and parameters are required arguments for the endpoint and run_description is optional.
  - PATCH to modify an existing run. Run_id is a required argument, run_name, run_description and parameters are optional argument and no other arguments are acepted.
  - DELETE to remove the run from the DB. Run_id is a required argument.
- Add new test for the run endpoints:
  - Create **_fixture_dummy_experiment_id_** to create a dummy experiment for the run tests.
  - The tests checks all the endpoints and its errors. They also check if the client wants to modify a "read-only" field like model_name.
- Change Run train_metrics, test_metrics, validation_metrics, start_time and end_time to nullable fields. Add default value to the status column (NOT_STARTED)


DashAI/back/api/api_v1/endpoints/runs.py
```diff
@@ -0,0 +1,207 @@
...
+logging.basicConfig(level=logging.DEBUG)
+log = logging.getLogger(__name__)
+
+router = APIRouter()
+
+
+@router.get("/")
+async def get_run(run_id: int, db: Session = Depends(get_db)):
+    """
+    Returns the run with the spécified id.
```
pbadillatorrealba:
```suggestion
    Returns the run with the specified id.
```

DashAI/back/api/api_v1/endpoints/runs.py
```diff
@@ -0,0 +1,207 @@
...
+
+    Returns
+    -------
+    JSON
+        run JSON
```
pbadillatorrealba: The docstring should indicate a python type (JSON is not) and a slightly more detailed description..

```suggestion
    dict
        A dict describing the requested run.
```

DashAI/back/api/api_v1/endpoints/runs.py
```diff
@@ -0,0 +1,207 @@
...
+            raise HTTPException(
+                status_code=status.HTTP_404_NOT_FOUND,
+                detail="Run not found",
+            )
+    except exc.SQLAlchemyError as e:
+        log.exception(e)
+        raise HTTPException(
+            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+            detail="Internal database error",
+        )
```
pbadillatorrealba: It is recommended that when you throw an exception in an exception handler (`except`) you use `from` to indicate that the exception you want to throw is caused by the exception caught (`e`) and not by the other code inside the except (such as `log.exception` or other).

```suggestion
        ) from e
```

pbadillatorrealba: Reference: https://stackoverflow.com/a/24752607

DashAI/back/api/api_v1/endpoints/runs.py
```diff
@@ -0,0 +1,207 @@
...
+
+    Parameters
+    ----------
+    experiment_id : int
+        id of the experiment assoc with the runs to query.
+
+    Returns
+    -------
+    List[JSON]
+        List of run JSON associated with the specified experiment id.
```
pbadillatorrealba:
```suggestion
    List[dict]
        A list of runs associated with the specified experiment id.
```

DashAI/back/api/api_v1/endpoints/runs.py
```diff
@@ -0,0 +1,207 @@
...
+from DashAI.back.database.models import Experiment, Run
+
+logging.basicConfig(level=logging.DEBUG)
+log = logging.getLogger(__name__)
+
+router = APIRouter()
+
+
+@router.get("/")
+async def get_run(run_id: int, db: Session = Depends(get_db)):
```
pbadillatorrealba: Small name change to make this function more descriptive.

```suggestion
async def get_run_by_id(run_id: int, db: Session = Depends(get_db)):
```

DashAI/back/api/api_v1/endpoints/runs.py
```diff
@@ -0,0 +1,207 @@
...
+        log.exception(e)
+        raise HTTPException(
+            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+            detail="Internal database error",
+        )
+    return run
+
+
+@router.get("/list/{experiment_id}")
+async def get_run_from_experiment(experiment_id: int, db: Session = Depends(get_db)):
```
pbadillatorrealba:
```suggestion
async def get_runs(experiment_id: int | None = None, db: Session = Depends(get_db)):
```

DashAI/back/api/api_v1/endpoints/runs.py
```diff
@@ -0,0 +1,207 @@
...
+    except exc.SQLAlchemyError as e:
+        log.exception(e)
+        raise HTTPException(
+            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+            detail="Internal database error",
+        )
+    return run
+
+
+@router.get("/list/{experiment_id}")
```
pbadillatorrealba: This endpoint is the one I commented that should be simply a get with an optional experiment_id parameter.
For this:

1. Change the path to `/`
2. Change the function name to `get_runs`
2. Add an optional `experiment_id` parameter that in case it is not null, return only the runs associated to the requested experiment.

DashAI/back/api/api_v1/endpoints/runs.py
```diff
@@ -0,0 +1,207 @@
...
+    mdoel_name : str
+        Name of the Model linked to the run.
+    name : str
+        Name of the run
+    parameters : JSON
+        Parameters of the run.
+    Returns
+    -------
+    JSON
+        JSON with the new run on the database
```
pbadillatorrealba:
```suggestion
    dict
        A dict with the new run on the database
```

DashAI/back/api/api_v1/endpoints/runs.py
```diff
@@ -0,0 +1,207 @@
...
+    db: Session = Depends(get_db),
+):
+    """
+    Endpoint to create run of an experiment.
+
+    Parameters
+    ----------
+    experiment_id : int
+        Id of the Experiment linked to the run.
+    mdoel_name : str
```
pbadillatorrealba:
```suggestion
    model_name : str
```

DashAI/back/api/api_v1/endpoints/runs.py
```diff
@@ -0,0 +1,207 @@
...
+    return runs
+
+
+@router.post("/", status_code=status.HTTP_201_CREATED)
+async def upload_run(
+    experiment_id: int,
+    model_name: str,
+    name: str,
+    parameters: dict,
+    description: str = "",
```
pbadillatorrealba: If empty, then shouldn't the default value be None?
(idk what happens with this change in the db).

```suggestion
    description: str = None,
```

tests/back/api/test_runs.py
```diff
@@ -0,0 +1,155 @@
...
+    data = response.json()
+    assert data["run_name"] == "Run2"
+
+
+def test_get_all_runs(client: TestClient, experiment_id: int):
+    response = client.get(f"/api/v1/run/list/{experiment_id}")
+    assert response.status_code == 200, response.text
+    data = response.json()
+    assert data[0]["experiment_id"] == experiment_id
+    assert data[1]["experiment_id"] == experiment_id
```
pbadillatorrealba: With the get functions change, this tests should be also changed.

pbadillatorrealba (changes requested):
Overall the implementation is good :)

However, I think there is a wrong design in the runs getters that needs to be changed prior merge.

- get to `/{id}` should return an specific run.
- get to `/` should return a list with all runs.
- get to `/` with the param `experiment_id` should return a list with the selected runs.

The first option should be implemented using a separate function (`get_run_by_id`) in the `/run/{id}` get .
On the other hand, the last two options should be implemented in the same function (`get_runs`), just by using or not the `experiment_id` parameter; in the `/run/` get path.

I left more implementation details in the methods plus another small code comment, most of them with suggestions.

Extra: this page has a good compilation of good api design: https://www.vinaysahni.com/best-practices-for-a-pragmatic-restful-api

pbadillatorrealba (approved):

PR #66, opened 2023-05-15
Added dummy frontend page and dummy backend endpoint. Removed legacy example_data

cristian-tamblay:
- You can now access a dummy endpoint at /app/dummy to start coding faster
- Removed example_data from frontend that was moved into json-server


DashAI/front/src/tabs/Dummy.jsx
```diff
@@ -0,0 +1,39 @@
+import React, { useState } from "react";
+import { getDummy as getDummyRequest } from "../api/dummy";
+import { StyledButton, P } from "../styles/globalComponents";
```
maximi4321:
```suggestion
import { Button, Typography } from "@mui/material";
```
I believe those components will be deprecated in the future,  the latest changes in the code use material-ui instead.

DashAI/front/src/tabs/Dummy.jsx
```diff
@@ -0,0 +1,39 @@
...
+    }
+  }
+
+  return (
+    <React.Fragment>
+      <div>
+        <StyledButton variant="dark" onClick={getDummy}>
+          Get Dummy
+        </StyledButton>
+        <P>Add your custom React code here</P>
```
maximi4321:
```suggestion
        <Button variant="outlined" onClick={getDummy}>
          Get Dummy
        </Button>
        <Typography variant="p">Add your custom React code here</Typography>
```

PR #67, opened 2023-05-15
Dataset modal - last step

maximi4321:
- The `ConfigureAndUploadDataset.jsx` component is added to combine the file upload and the dataloader parameters configuration in the final step.
- The `ConfigurableObject/ParamsModal.jsx` is refactored to display its content without the need for a modal. It is also adapted to follow the dataset modal flow. Therefore, it is renamed and moved to end up as: `datasets/DataloaderConfiguration.jsx`.
- The `Upload.jsx` is refactored. It no longer communicates with the API in this component, it only modifies a state to save the dataset.
- A small modification is made in `ParameterForm.jsx` and `MainForm.jsx` to allow submitting the form with external buttons to these components.


pbadillatorrealba:
Please resolve conflicts before proceed!

Also, I could not run the process until I reached the end. Could you tell me how to test it?
Do you use yarn json-server or the backend for this? How?

maximi4321:
> Please resolve conflicts before proceed!

The conflicts are resolved, but with the merge I had to make a small change in `FormRenderer.jsx` so that in the case of the "list_of_strings" type it returns a `<div />`. Creating a component to handle that type correctly is left for another PR.

> Do you use yarn json-server or the backend for this? How?

I used the backend since it already had the endpoints i needed. To use json-server, I would need to add nested routes such as "dataloaders/TabularClassificationTask" and some old endpoints (v0).

> Could you tell me how to test it?

2. Go back to the root folder `DashAI/` and run `python -c "import DashAI;DashAI.run()"`.
3. Access localhost:8000/app.
4. Navigate to "Dataset" tab and click the "New Dataset" button
5. Choose the task "Tabular Classification" (the other tasks don't have compatible dataloaders)
6. Choose the dataloader "CSV Data" ("JSON Data" has a problem in the upload)
7. In the last step, upload the file`DashAI/DashAI/back/example_datasets/iris.csv`
8. You can name the dataset (or leave blank) and press save (you can also configure the parameters, but since there is no validation it is easy to get an error)

This .gif is an example of how to test it.
upload-dataset

pbadillatorrealba (approved):
The code in general looks good.
Unfortunately because of the registry APIs change I couldn't test it, but I accept it to keep us moving forward.
Later we can make bugfixes.

PR #68, opened 2023-05-18
Experiment modal - last step

maximi4321:
# Summary

Last step of the modal to create a new experiment, in this step models can be added and configured.

## Type of change

- Front end new feature.

## Changes

- The `ConfigureModelsStep.jsx` component is created as a container for the logic for the final step 
- A table (`ModelsTable.jsx`) is added to: 
  - display models that are currently in the experiment 
  - configure the parameters of a model (`EditModelDialog.jsx`) 
  - remove models from the experiment (`DeleteModelDialog.jsx`) 
- Some modifications are made to `ParameterForm.jsx` so that it can receive external initial values and thus be able to load the user's configurated values after having closed the configuration modal  
- The request to the experiment endpoint is updated 
- The `ClassInput.jsx` component is modified to support two options for recursive parameters: Modal or Collapsible

## How to Test

Option 1 - Backend
3. Navigate to the experiments tab
4.  Click the "New Experiment" button
5. Create a name and select a task for your experiment
6. Select the dataset for your experiment
7. Add a model to the table, optionally configure it and save the experiment.
8. Click "Update" and the new experiment should be visible in the table

## Screenshots

[createExperiment.webm](https://github.com/DashAISoftware/DashAI/assets/53657198/2e18d67d-4e7e-451a-b8b3-0c9fc9c56852)

### DIFF
--- a/DashAI/front/src/api/experiment.ts
+++ b/DashAI/front/src/api/experiment.ts
@@ -12,19 +12,8 @@ export const getExperimentById = async (id: string): Promise<IExperiment> => {
   return response.data;
 };
 
-// export const createExperiment = async (formData: FormData): Promise<object> => {
-//   const response = await api.post<IExperiment>("/v1/experiment", formData);
-//   return response.data;
-// };
-
-export const createExperiment = async (
-  datasetId: number,
-  taskName: string,
-  name: string,
-): Promise<IExperiment> => {
-  const response = await api.post<IExperiment>(
-    `${endpointURL}/?dataset_id=${datasetId}&task_name=${taskName}&name=${name}`,
-  );
+export const createExperiment = async (formData: FormData): Promise<object> => {
+  const response = await api.post<IExperiment>("/v1/experiment", formData);
   return response.data;
 };
 
--- a/DashAI/front/src/api/experiment.ts
+++ b/DashAI/front/src/api/experiment.ts
@@ -1,7 +1,7 @@
 import api from "./api";
 import type { IExperiment } from "../types/experiment";
 
-const endpointURL = "/v1/experiment"
+const endpointURL = "/v1/experiment";
 export const getExperiments = async (): Promise<IExperiment[]> => {
   const response = await api.get<IExperiment[]>(endpointURL);
   return response.data;
@@ -17,10 +17,16 @@ export const getExperimentById = async (id: string): Promise<IExperiment> => {
 //   return response.data;
 // };
 
-export const createExperiment = async (datasetId: number, taskName: string, name: string ): Promise<IExperiment> => {
-  const response = await api.post<IExperiment>(`${endpointURL}/?dataset_id=${datasetId}&task_name=${taskName}&name=${name}`)
-  return response.data
-}
+export const createExperiment = async (
+  datasetId: number,
+  taskName: string,
+  name: string,
+): Promise<IExperiment> => {
+  const response = await api.post<IExperiment>(
+    `${endpointURL}/?dataset_id=${datasetId}&task_name=${taskName}&name=${name}`,
+  );
+  return response.data;
+};
 
 export const updateExperiment = async ({
   id,
--- a/DashAI/front/src/api/experiment.ts
+++ b/DashAI/front/src/api/experiment.ts
@@ -1,20 +1,26 @@
 import api from "./api";
 import type { IExperiment } from "../types/experiment";
 
+const endpointURL = "/v1/experiment"
 export const getExperiments = async (): Promise<IExperiment[]> => {
-  const response = await api.get<IExperiment[]>("/v1/experiment");
+  const response = await api.get<IExperiment[]>(endpointURL);
   return response.data;
 };
 
 export const getExperimentById = async (id: string): Promise<IExperiment> => {
-  const response = await api.get<IExperiment>(`/v1/experiment/${id}`);
+  const response = await api.get<IExperiment>(`${endpointURL}/${id}`);
   return response.data;
 };
 
-export const createExperiment = async (formData: FormData): Promise<object> => {
-  const response = await api.post<IExperiment>("/v1/experiment", formData);
-  return response.data;
-};
+// export const createExperiment = async (formData: FormData): Promise<object> => {
+//   const response = await api.post<IExperiment>("/v1/experiment", formData);
+//   return response.data;
+// };
+
+export const createExperiment = async (datasetId: number, taskName: string, name: string ): Promise<IExperiment> => {
+  const response = await api.post<IExperiment>(`${endpointURL}/?dataset_id=${datasetId}&task_name=${taskName}&name=${name}`)
+  return response.data
+}
 
 export const updateExperiment = async ({
   id,
--- a/DashAI/front/src/components/experiments/ConfigureModelsStep.jsx
+++ b/DashAI/front/src/components/experiments/ConfigureModelsStep.jsx
@@ -1,17 +1,24 @@
-import React, { useState } from "react";
+import React, { useEffect, useState } from "react";
 import PropTypes from "prop-types";
 import { Button, Grid, MenuItem, TextField, Typography } from "@mui/material";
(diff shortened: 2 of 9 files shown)


DashAI/front/src/api/experiment.ts
```diff
@@ -1,18 +1,30 @@
 import api from "./api";
 import type { IExperiment } from "../types/experiment";
 
+const endpointURL = "/v1/experiment";
```
pbadillatorrealba, line 4: Perhaps it would be good to re-centralize this type of configuration variables in a `.env` file. What do you think?

DashAI/front/src/api/experiment.ts
```diff
@@ -1,18 +1,30 @@
...
+// };
+
+export const createExperiment = async (
+  datasetId: number,
+  taskName: string,
+  name: string,
+): Promise<IExperiment> => {
+  const response = await api.post<IExperiment>(
+    `${endpointURL}/?dataset_id=${datasetId}&task_name=${taskName}&name=${name}`,
+  );
```
pbadillatorrealba: As a general rule, in a post the parameters should be placed in the body of the request, not as a query in the url.


```suggestion
    const response = await api.post<IExperiment>(endpointURL, {
        dataset_id: datasetId,
        task_name: taskName,
        name: name
        })
```

If there is a problem with the back compatibility, the back API should be changed.

DashAI/front/src/components/experiments/ConfigureModelsStep.jsx
```diff
@@ -0,0 +1,182 @@
...
+  const [selectedModel, setSelectedModel] = useState("");
+  // const [modelsInExperiment, setModelsInExperiment] = useState([]);
+
+  const compatibleModels = [
+    "KNeighborsClassifier",
+    "RandomForestClassifier",
+    "SVC",
+    "numericalwrapperfortext",
+    "tcTransformerEngSpa",
+  ];
```
pbadillatorrealba, line 28: This will be changed in a near future? If this is the case, then add a card to notion to keep tracked this issue.

DashAI/front/src/components/experiments/ConfigureModelsStep.jsx
```diff
@@ -0,0 +1,182 @@
...
+
+  // width for model nickname and model type textfields
+  const textFieldWidth = "32vw";
+
+  // gets a new id for a model by adding 1 to the max id of the current models in the experiment
```
pbadillatorrealba: Each model will map to a unique run right?
Then, why not to use an [UUID](https://interactivechaos.com/es/wiki/identificador-unico-universal-uuid)?
Here are a simple reference of this: [UUID in JS - Stackverflow](https://stackoverflow.com/questions/105034/how-do-i-create-a-guid-uuid)

DashAI/front/src/components/experiments/ConfigureModelsStep.jsx
```diff
@@ -0,0 +1,182 @@
...
+  }, [newExp]);
+
+  // fetches the JSON object of the model selected by the user
+  // useEffect(() => {
+  //   if (selectedModel !== "") {
+  //     getObjectSchema();
+  //   }
+  // }, [selectedModel]);
+
+  // console.log(objectSchema);
```
pbadillatorrealba: Delete unused code:

```suggestion
```

DashAI/front/src/components/experiments/EditModelDialog.jsx
```diff
@@ -0,0 +1,124 @@
+import React, { useEffect, useState } from "react";
+import PropTypes from "prop-types";
+import { GridActionsCellItem } from "@mui/x-data-grid";
+import EditIcon from "@mui/icons-material/Edit";
+import ParameterForm from "../ConfigurableObject/ParameterForm";
+import {
+  // Button,
```
pbadillatorrealba:
```suggestion
```

DashAI/front/src/components/experiments/EditModelDialog.jsx
```diff
@@ -0,0 +1,124 @@
...
+  CircularProgress,
+  Dialog,
+  // DialogActions,
```
pbadillatorrealba:
```suggestion
```

DashAI/front/src/components/experiments/EditModelDialog.jsx
```diff
@@ -0,0 +1,124 @@
...
+              )}
+            </Grid>
+          </Paper>
+        </DialogContent>
+        {/* <DialogActions>
+          <Button onClick={() => setOpen(false)} autoFocus>
+            Cancel
+          </Button>
+          <Button onClick={handleEdit}>Save</Button>
+        </DialogActions> */}
```
pbadillatorrealba: Delete if it is unused.
```suggestion
```

DashAI/front/src/components/experiments/ModelsTable.jsx
pbadillatorrealba, line 1: Maybe is this file still in development?
I will skip it.

DashAI/front/src/components/experiments/NewExperimentModal.jsx
```diff
@@ -40,11 +45,39 @@ const defaultNewExp = {
...
+
+      if (error.response) {
+        console.error("Response error:", error.message);
+      } else if (error.request) {
+        console.error("Request error", error.request);
+      } else {
+        console.error("Unkown Error", error.message);
+      }
+    } finally {
+      // setLoading(false);
```
pbadillatorrealba: The [MUI button has a property to set it as loading](https://mui.com/material-ui/react-button/#loading-button). It is nice and keeps the user informed with respect to the request status.
I think that is great idea to replace the original button with the previous one.

pd: Sorry, i don't know where was the propper place to put this message than here ;)

pbadillatorrealba (changes requested):
As the previous pr, I could not open the interface due the registry API changes.
But I left some comments with respect to the API calls, component suggestions and commented code.

DashAI/front/src/api/experiment.ts
maximi4321: I think it would be similar when using the backend, with the difference that all the urls would be in a single file.

The problem, I think, is that the variables would need to be duplicated in both `.env.development` and `.env.production`. However, if we want to allow them to have different values in each environment, I think it would be a good flexibility improvement.

DashAI/front/src/api/experiment.ts
maximi4321: I restored the previous `createExperiment` function that receives a `formData` as an argument.

>If there is a problem with the back compatibility, the back API should be changed.

Yes, there are compatibility issues with the backend.
I have been testing some things, and it can be fixed with a small modification to the `upload_experiment` endpoint in `DashAI/back/api/api_v1/endpoints/experiments.py`.

modifying
```
...
async def upload_experiment(
    dataset_id: int,
    task_name: str,
    name: str,
    db: Session = Depends(get_db),
):
...
```
to
```
...
async def upload_experiment(
    dataset_id: int = Form(...),
    task_name: str = Form(...),
    name: str = Form(...),
    db: Session = Depends(get_db),
):
...
```

I'm not sure if there's a better solution or whether to upload that change here. Maybe it's better to create another PR.

cristian-tamblay: Lets resolve this in a new pr in the future

cristian-tamblay:
I had a problem with the responsiveness on the fullscreen version on a 1080p monitor. The "Add" button is weird. Fix this and its ready to approve.
image

maximi4321:
> I had a problem with the responsiveness on the fullscreen version on a 1080p monitor. The "Add" button is weird. Fix this and its ready to approve.

In the latest commit i added some Grid components to improve responsiveness.

DashAI/front/src/api/experiment.ts
pbadillatorrealba: Ok, let's put it in tasks for the future.

pbadillatorrealba (approved):

PR #70, opened 2023-05-19
Improve backend code quality

pbadillatorrealba:
# Summary

This pr improves the quality while avoiding potential bugs of the backend code through the revision of various [ruff](https://github.com/charliermarsh/ruff) linter rules.
Given the latest syntax changes, this pr requires `Python 3.8` to operate properly. 
Also, it requires `ruff>0.0.270` to obtain the same set of used rules.

## Type of change

- Bug fix.
- Refactoring.

## Changes

- Update ruff rules, disabling some unuseful. Some annoying ones were disabled, while new ones were enabled. Check `pyproject.toml` to see ruff enabled rules.
- Change class methods first input from `self` to `cls` (because a class method recieves as a input the class, not the instantiated object).
- Update types definition to python 3.8 specifications.
- Add `from e` to every exception raised (`raise SomException() from e`) inside other exception in a `except` block. This makes the exception displayed by the interpreter can be distinguished between the one you threw versus the one that may have been accidentally thrown by a code error in the except.
- Add several trailing commas in function arguments to re-format the code to improve code legibility (trailing commas cause arguments to be automatically separated into new lines).
- Add context managers to every ill used `open()` in the code.
- Change string contatenations from explicit (i.e., using `+`) to implicit (without `+`, only using newlines).
- Improve several docstrings descriptions.
- Improve several docstring formats with the D ruff rules.
- Update several `pytest.exception` context managers without a match.
- Delete `ABCMeta`from ConfigurableObject (since there is no abstract method to implement).
- Improved some exception messages.
- Delete useless `.keys()` in places where dictionary keys are iterated.
- Merge separated `isinstance` into one. For example `(isinstance(class_column, str) or isinstance(class_column, int))` to `isinstance(class_column, (str, int))`.
- The registers were modified to be compatible with the syntax supported by Python 3.8.

The pr also deletes a `test_sample.py` test file which contained a sample test file not applicable to the project.

## How to Test

No new tests were implemented because no new features were implemented, only bugfixes and refactoring.
Usual: use `pytest .` to check all backend tests.

## Notes

- `flake8` continues to be the default linter for the back and pre-commit, as there are several problems detected by ruff yet to be resolved.
- Depends on #80.


DashAI/back/dataloaders/classes/dataset_dashai.py
```diff
@@ -119,33 +111,35 @@ def load_from_disk(dataset_path: str):
...
+            if column in columns:
                 pass
             else:
-                raise ValueError(f"Class column '{c}' does not exist in dataset.")
+                raise ValueError(f"Class column '{column}' does not exist in dataset.")
         # set columns types
         new_features = self.features.copy()
-        for c in types.keys():
+        for column in types:
             # ESTO ES TEMPORAL
```
inunezn7:
```suggestion
            # this is temporary
```

DashAI/back/dataloaders/classes/dataset_dashai.py
```diff
@@ -119,33 +111,35 @@ def load_from_disk(dataset_path: str):
...
 
-        for c in types.keys():
-            if c in columns:
+        for column in types:
+            if column in columns:
                 pass
             else:
-                raise ValueError(f"Class column '{c}' does not exist in dataset.")
+                raise ValueError(f"Class column '{column}' does not exist in dataset.")
         # set columns types
```
inunezn7, line 135: I think this comment is unnecessary

DashAI/back/models/classes/getters.py
```diff
@@ -38,10 +38,7 @@ def filter_by_parent(parent_class_name):
...
 
 def get_model_params_from_task(task_name):
-    """
-    It returns a dictionary with a list of the available models for the
-    task or an error if the squema
-    """
+    """Return a dictionary with a list of the available models for the task."""
     model_class_name = f"{task_name[:-4]}Model"
     try:
         dict = filter_by_parent(model_class_name)
```
inunezn7, line 44: I think we can delete this method, it's really old and it is not used anywhere. We don't even use it in the old api version

DashAI/back/models/classes/tc_transformer_eng_spa.py
```diff
@@ -50,39 +50,31 @@ def __init__(self, **kwargs) -> None:
...
         )
         self.data_collator = DataCollatorForSeq2Seq(self.tokenizer, model=self.model)
         self.metric_bleu = evaluate.load("sacrebleu")
         self.metric_ter = evaluate.load("ter")
 
     def tokenize_input(self, d):
-        """
-        Function to tokenize the input, d should be a DatasetDict
-        """
+        """Tokenize the input, d should be a DatasetDict."""
```
inunezn7, line 60: What should be the convention here? For what I've seen before shouldn't be something like "d: DatasetDict" in a different line?

DashAI/back/tasks/text_classification_task.py
```diff
@@ -72,7 +68,7 @@ def validate_dataset(self, dataset: DatasetDict, class_column: Union[str, int]):
             elif "string" not in data_type:
                 return (
                     "Dataset have non-text data. "
-                    + f"Make sure you have only text for {self.NAME}."
+                    f"Make sure you have only text for {self.NAME}."
```
inunezn7: This should be _ _ name _ _ instead of NAME, right?

inunezn7 (approved):
Great standardization of the code. I left some small comments

DashAI/back/dataloaders/classes/dataset_dashai.py
pbadillatorrealba: Which comment?

`# set column types`?

DashAI/back/models/classes/getters.py
pbadillatorrealba: Perhaps we could remove the entire module?

I don't think any of the methods in this file are being used.

DashAI/back/models/classes/tc_transformer_eng_spa.py
pbadillatorrealba: That's right.
However, I think that @Oziel14  could improve these docstrings since at first glance I don't have much knowledge of the inputs, outputs or what they do.

DashAI/back/dataloaders/classes/dataset_dashai.py
inunezn7: yeah, that one. Just a suggestion, please feel free to keep it if you considere it necessary

DashAI/back/tasks/text_classification_task.py
pbadillatorrealba: Ok!

DashAI/back/models/classes/getters.py
inunezn7: the method filter_by_parent is used in the old version of the api. Maybe we should remove the old api version and related modules in a different PR

DashAI/back/models/classes/tc_transformer_eng_spa.py
Oziel14: Like the previous comment, this class is from the old version and will change completely, so it is not worth commenting on it. We might as well delete it.

cristian-tamblay (approved):

PR #63, opened 2023-05-11
Validators and task changes

Oziel14:
This PR depends on #60 

Summary List:

- Validators have been added to the basetask class, which checks that the types and cardinality of the columns that act as input or output are correct, also a method called prepare to task has been added that transforms some columns to the types needed by the task.
- To the tabular classification task the types and cardinalities compatible with the validator were added, and the corresponding changes were made in prepare for tasks to transform its labels column to the corresponding type.
- Tests were also carried out to test some of the mentioned functionalities.

PR #71, opened 2023-05-24
Modification in save and load methods for datasets

Oziel14:
# Summary
The load function has been changed and now it can receive a complete datasetdict instead of a dashaidataset, like the previous one, it is enough to receive the path where the dataset is stored.

A function called save_dataset was made which receives a datasetdict and the path where it should be saved, this way it avoids the code to save each dashaidataset one by one, it uses the save_from_disk method of the DashAIDataset class.

## Type of change

- Refactoring.
- Feature

## Changes

- Refactoring 1
- Feature 1

## How to Test
Inside the test_dashai_dataset.py there is a function called test_save_to_disk_and_load to test the new load and save, thus ensuring that you can load and read datasetdicts with dashaidatasets inside.

## Notes
This PR depends on the PR #63


tests/back/dataloaders/test_dashai_dataset.py
```diff
@@ -147,24 +147,17 @@ def test_save_to_disk_and_load():
...
-    )
+    save_dataset(separate_dataset, "tests/back/dataloaders/dashaidataset")
+
+    # for i in separate_dataset.keys():
+    #     separate_dataset[i].save_to_disk(f"tests/back/dataloaders/dashaidataset_{i}")
+    # paths = {
+    #     "path_train": "tests/back/dataloaders/dashaidataset_train",
+    #     "path_val": "tests/back/dataloaders/dashaidataset_validation",
+    #     "path_test": "tests/back/dataloaders/dashaidataset_test",
+    # }
```
pbadillatorrealba: Delete commented code when you are pushing to staging. 
Otherwise, we tend to fill up with commented code that we don't know what it does or why it is there 😥.
```suggestion
```

pbadillatorrealba (changes requested):

tests/back/dataloaders/test_dashai_dataset.py
Oziel14: My bad, i forgot to delete it

pbadillatorrealba (approved):

PR #72, opened 2023-05-24
Runner Endpoint

Riul1999:
# Summary

- Runner endpoint (path _/runner/?run_id=_).
- Test test_runner.
- Run endpoint: Delete trained run file from file system.
- Run model: Change run_name and run_description to name and description.
- Test Dataset: Delete all the created dataset.

This PR depends on https://github.com/DashAISoftware/DashAI/pull/78

## Type of change

- Back end new feature.

## Changes

Add runner endpoint (_/runner/_) to the app. It's only available the post method which train and test the specified run.

The runner endpoint follow this pipeline:
- Get the Run, Experiment and Dataset from the DB.
- Instantiate the run with it's parameters.
- Instantiate the Dataset. 
- Train the run.
- Get the train and test results using the metrics associated with the Task associated to the Experiment. (TBD)
- Save the trained run.

Add test _test_runner_ to check the behavior of the endpoint.
- The test depends on dataset endpoint.
- The test mocks the registry and use own Dummy components.

Run endpoint:
- If the run was trained (i.e. RunStatus = FINISHED) tries to delete the saved run file.
- If the run was trained and the file does not exists it throws a HTTP Exception.

Run database model: 
- Change run_name and run_description to name and description.
- Refactor the change in the Run endpoint and in the test_runs files.

Test Dataset: 
- Delete all the created dataset. The remaining datasets could cause errors in other tests that uses datasets (like test_runner)

## How to Test

In the root of the proyect you can run `pytest .\tests\back\api\test_runner.py ` to execute the test.
It's also good idea to run `pytest .` to check test_runs and test_datasets are still working.


DashAI/back/api/api_v1/endpoints/runner.py
```diff
@@ -0,0 +1,99 @@
...
+        if not run:
+            raise HTTPException(
+                status_code=status.HTTP_404_NOT_FOUND, detail="Run not found"
+            )
+        exp: Experiment = db.get(Experiment, run.experiment_id)
+        if not exp:
+            raise HTTPException(
+                status_code=status.HTTP_404_NOT_FOUND, detail="Experiment not found"
+            )
+        dat: Dataset = db.get(Dataset, exp.dataset_id)
```
pbadillatorrealba: Try to use the complete name to facilitate code reading (change everywhere else)

```suggestion
        dataset: Dataset = db.get(Dataset, exp.dataset_id)
```

DashAI/back/api/api_v1/endpoints/runner.py
```diff
@@ -0,0 +1,99 @@
...
+            raise HTTPException(
+                status_code=status.HTTP_404_NOT_FOUND, detail="Dataset not found"
+            )
+    except exc.SQLAlchemyError as e:
+        log.exception(e)
+        raise HTTPException(
+            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+            detail="Internal database error",
+        )
+    run_instance = component_registry[run.model_name]["class"](**run.parameters)
```
pbadillatorrealba: This line of code is too complex: it has too many operations.
Split at least in two:

```suggestion
    # Instance the model with the run parameters.
    run_model_class = component_registry[run.model_name]["class"]
    run_instance = run_model_class(**run.parameters)
```

DashAI/back/core/config.py
```diff
@@ -36,6 +36,7 @@ class Settings(BaseSettings):
     DB_PATH: str = "DashAI/back/database/DashAI.sqlite"
     FRONT_BUILD_PATH: str = "DashAI/front/build"
     USER_DATASET_PATH: str = "DashAI/back/user_datasets"
+    USER_RUN_PATH: str = "DashAI/back/user_runs"
```
pbadillatorrealba, line 39: Maybe to discuss in the future, but wouldn't it be nice to save the user's artifacts in the user's home? The user will probably never find the path where the package is installed.

DashAI/back/database/models.py
```diff
@@ -64,11 +64,24 @@ class Run(Base):
     train_metrics: Mapped[JSON] = mapped_column(JSON, nullable=True)
     test_metrics: Mapped[JSON] = mapped_column(JSON, nullable=True)
     validation_metrics: Mapped[JSON] = mapped_column(JSON, nullable=True)
+    # artifacts
+    # artifacts: Mapped[str] = mapped_column(JSON)
```
pbadillatorrealba: Why to delete the artifacts?

pbadillatorrealba (changes requested):
This implementation looks good :)

I left some style comments in the code. I also have two questions:

1. Wouldn't it be better if this APIs file only handles the networking and the logic is elsewhere, for example in a file in back/core/runner.py?

2. What if it fails to get the classes, or the training? Or calculating metrics, or saving the file? Maybe we should throw custom exceptions for any of these cases or return an HTTPException? Perhaps this point can go to the backlog in the meantime.


Remember to update the pr description and change its destination to staging.

cristian-tamblay (approved):

pbadillatorrealba (approved):

PR #75, opened 2023-05-24
Modify ItemsList.jsx to make it more reusable

maximi4321:
# Summary

Change `ItemsList.jsx` to `ItemSelector.jsx` which is a more general version for reuse in other components, these modifications also allow to keep the values selected by the user in case he goes back to previous steps in the modal.

## Type of change

- Front end new feature.
- Refactoring

## Changes

- Created the `ItemSelector.jsx` component as a general version of a list of selectable items that displays a description of the selected item.
- Created `SelectTaskStep.jsx` and `SelectDataloaderStep.jsx` as steps of the dataset modal that use `ItemSelector.jsx`
- Added as a new feature the ability to keep the user's selection even if the user changes the step.

## How to Test

3. Navigate to the datasets tab and create a new dataset by clicking the "New Dataset" button
4.  Select one of the tasks
5. Select one of the dataloaders
6. Navigate back to the step where a task is selected and you will see that the task you selected earlier is still selected.

## Screenshots

itemSelector (1)

## Notes

This PR depends on #73


pbadillatorrealba (commented):
The refactoring is really nice!

I would recommend leaving these components in a separate folder, something like components/custom or components/internal.

Question: I think it is similar to the one in the experiment modal. could it be changed also in this same pr?

maximi4321:
> Question: I think it is similar to the one in the experiment modal. could it be changed also in this same pr?

Yes, I believe it is possible.
But this PR requires more work, now in the results tab we need a similar functionality but without the part of the selected item description.

pbadillatorrealba:
@maximi4321 , please resolve conflicts before merge!

pbadillatorrealba (approved):

PR #76, opened 2023-05-24
Edit dataset

maximi4321:
# Summary

Includes a modal to change the name and task associated with a dataset

## Type of change

- Front end new feature.

## Changes

- Add a modal with a TextField to change the name of a dataset and also renders `ItemSelector.jsx` to change the task associated to the dataset
- updates the Datasets table once the update is done

## How to Test

Option 1 - Backend
3. Navigate to the datasets tab
4.  Click the button to edit dataset (requires at least one dataset in the table)
5. Modify the name and the task and click "Save"
6. The modified values should be updated in the datasets table.

Option 2 - json-server
1. Go to `DashAI/DashAI/front` and run `yarn json-server`
2. in another terminal in the same folder of step 1 run `yarn start`
4. Navigate to the datasets tab
5.  Click the button to edit dataset (requires at least one dataset in the table)
6. Modify the name and the task (this option doesn't support updating the values in the table)

## Screenshots

edit-dataset

## Notes

This PR depends on #75


pbadillatorrealba (changes requested):
I think the idea of editing tasks is good but I strongly recommend disabling it. Next I explain why:

Changing the name is simple, just modify the name in the DB and that's it; now all objects that depend on the dataset will see the new name (assuming the id will remain the same).

However, changing the task would mean that all the experiments that depend on this dataset would have none associated with it. In that case, actions must be taken to prevent this from happening. Therefore, the easiest thing to do is to disable editing.

Interesting note: we should also think about what to do when deleting a dataset in the future. I am of the idea of only allowing the deletion of datasets that do not have any associated experiment.

DashAI/front/src/components/custom/ItemSelector.jsx
```diff
@@ -0,0 +1,188 @@
+import React, { useEffect, useState } from "react";
+import {
+  Grid,
+  // DialogContentText,
```
pbadillatorrealba:
```suggestion
```

DashAI/front/src/components/EditDatasetModal.jsx
```diff
@@ -1,18 +1,175 @@
-import React from "react";
+import React, { useEffect, useState } from "react";
+import PropTypes from "prop-types";
 import { GridActionsCellItem } from "@mui/x-data-grid";
 import EditIcon from "@mui/icons-material/Edit";
+import {
+  Button,
+  // ButtonGroup,
```
pbadillatorrealba:
```suggestion
```

DashAI/front/src/api/datasets.ts
```diff
@@ -13,7 +13,18 @@ export const getDatasets = async (): Promise<IDataset[]> => {
...
 
+export const updateDataset = async (
+  id: number,
+  name: string,
+  taskName: string,
+): Promise<IDataset> => {
+  const response = await api.patch<IDataset>(
+    `${datasetEndpoint}/${id}?name=${name}&task_name=${taskName}`,
+  );
+  return response.data;
```
pbadillatorrealba: As convention, POST, PUT, PATCH should deliver the new data in the request body.
On the other hand, when the resource is already created and you want to modify it (in PUT and PATCH), it must be accessed through Paths (i.e., datasets/1, etc.).

To get an idea, this could be read as: 
_Update the dataset with id=1 with the json data provided in the body._

```suggestion
  const response = await api.patch<IDataset>(
    `${datasetEndpoint}/${id}`,
    {"name": name, "task_name":  task_name}
  );
  return response.data;
```

pbadillatorrealba (changes requested):
As I understood, you only disabled task editing, but did not delete it?

I left a last API design comment. It's pretty brief, but probably also requires a change in the back end.

DashAI/front/src/api/datasets.ts
maximi4321: Thank you!

I fixed it by using the same method as for solving the post request in the case of uploading an experiment, passing the elements of the object as formData.

The modifications on the backend to make it work are similar: add `Form(...)` to each parameter that requires it.

pbadillatorrealba (approved):
Approved!

PR #77, opened 2023-05-25
Train and predict Sklearn classification models

Oziel14:
# Summary
The necessary changes were made to train and predict in sklearn classification models, especially for the tabular classification task.

## Type of change
- Refactoring.

## Changes

- Added a method to format the data for the sklearn models that can be used, i.e. knn, smv and random forest for classification.
- The prediction models return probabilities instead of the corresponding label (see notes).
- Tests were done for training, prediction of a model, as well as the ability to save and load the model. 

## How to Test

There are tests in the file test_tabular_class_models.py

## Notes

According to what we discussed with Felipe, we are looking for a classification model to return the probabilities of each class instead of the class itself, in order to be able to use metrics that use this information.


pbadillatorrealba (changes requested):
I have been wondering for some time whether this type of method should be in a dataloader and not in the models. 

Why: Because the models should only know how to receive the data, train and predict, not process the data.

I think this part is more associated with the classes that intrinsically handle data, the DataLoaders.

Now, if you move that way, the question is who executes them.
From what I saw in the tests, at this point the execution is manual, you have to instantiate an object that inherits scikit-learn and then do format_data. 

I think it would be much better if the dataloader is asked to do something like that:
json_dataloader.to_pandas() or csv_dataloader.to_scikit() or something like that.

We should discuss this in the weekly meeting.

Extra note: you should delete all json and arrow files before to carry out the merge.

pbadillatorrealba (changes requested):
@Oziel14 , please delete all test arrow and json files before proceed with the merge.

pbadillatorrealba (approved):

PR #78, opened 2023-05-26
Metrics for classification tasks

Oziel14:
# Summary
This branch depends on the PR #77 .
The metrics class was created as well as a register for them.
Incorporated accuracy, recall, precision and f1score metrics.
Tests were made for the metrics

## Type of change
- Back end new feature.

## Changes
- Changes in core/config files

## How to Test
Tests were performed on the test_classification_metrics.py

## Notes
By approving this PR you would already have a complete back end pipeline to train and evaluate a model. But only the task of tabulating classification, sklearn models and the mentioned metrics.


DashAI/back/metrics/Classification/f1score.py
```diff
@@ -0,0 +1,48 @@
...
+        float
+            f1 score between true labels and predicted labels
+        """
+        validate_inputs(true_labels, probs_pred_labels)
+        pred_labels = np.argmax(probs_pred_labels, axis=1)
+        multiclass = len(np.unique(true_labels)) > 2
+        if multiclass:
+            return f1_score(true_labels, pred_labels, average="micro")
+        else:
+            return f1_score(true_labels, pred_labels, average="micro")
```
pbadillatorrealba: Both branches yields the same results! (check the average parameter).
The same applies for the rest of the metrics.

Note that there are also several ways to average by having multi-classes.
See more in the [f1 documentation](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.f1_score.html).

DashAI/back/metrics/Classification/accuracy.py
```diff
@@ -0,0 +1,44 @@
...
+
+    Parameters
+    ----------
+    true_labels : list
+        True labels
+    pred_labels : list
+        Predict labels by the model
+    """
+    if len(true_labels) != len(pred_labels):
+        raise ValueError("The length of the true and predicted labels must be equal.")
```
pbadillatorrealba: This function is often repeated. Why not put it in `ClassificationMetric` and then have the subclass simply inherit it?

DashAI/back/metrics/Classification/f1score.py
```diff
@@ -0,0 +1,48 @@
+import numpy as np
+from sklearn.metrics import f1_score
+
+from DashAI.back.metrics.classification_metric import ClassificationMetric
+
+
+class F1Score(ClassificationMetric):
```
pbadillatorrealba: `F1` or `F1Score` ?

The following metrics does not have that suffix.
I think it is good to standardize this at this point.

DashAI/back/metrics/Classification/precision.py
```diff
@@ -0,0 +1,48 @@
...
+        float
+            Precision score between true labels and predicted labels
+        """
+        validate_inputs(true_labels, probs_pred_labels)
+        pred_labels = np.argmax(probs_pred_labels, axis=1)
+        multiclass = len(np.unique(true_labels)) > 2
+        if multiclass:
+            return precision_score(true_labels, pred_labels, average="micro")
+        else:
+            return precision_score(true_labels, pred_labels, average="micro")
```
pbadillatorrealba: Same error as before.

DashAI/back/metrics/Classification/recall.py
```diff
@@ -0,0 +1,48 @@
...
+        float
+            recall score between true labels and predicted labels
+        """
+        validate_inputs(true_labels, probs_pred_labels)
+        pred_labels = np.argmax(probs_pred_labels, axis=1)
+        multiclass = len(np.unique(true_labels)) > 2
+        if multiclass:
+            return recall_score(true_labels, pred_labels, average="micro")
+        else:
+            return recall_score(true_labels, pred_labels, average="micro")
```
pbadillatorrealba: Same error as before.

pbadillatorrealba (changes requested):
I am in doubt.

Should we treat the metric as a configurable object?
Or a static one, like the tasks?

I ask because for the case of precision, recall and f1 in multiclass configuration have a parameter to configure how the average of the different values of the metric per class (micro, macro, weighted, etc...) will be calculated.

I particularly like the first option, but for simplicity, I would go for the second one and always keep the calculation fixed, making different F1 classes such as F1_micro, F1_macro, F1_weighted, etc...

What do you think of the options?

Another issue: seeing that the registry is close to being approved, merge the current metrics registry could incur in bugs. 
Would you be okay if you wait for the new log to be merged before merging this branch?

Oziel14:
About the first remark the idea is that it is a configurable object, in fact the base class of metrics comes from the class of configurable objects. But due to the time for now it will be static. And that explains that metrics like precision or recall for multiclasses use micro average.

And about the second point, there is no problem with waiting for the change of the registry before merging this branch

pbadillatorrealba (approved):
Looks great!

cristian-tamblay (changes requested):
We should change `DashAI/back/metrics/Classification` folder name to `DashAI/back/metrics/classification` so we don't have to fix it in the future.

Oziel14:
Name change done

cristian-tamblay (approved):

PR #80, opened 2023-05-29
Registries V2

pbadillatorrealba:
# Summary

New implementation of records and API serving components according to the new single registry definition plus unrestricted relationships.

## Type of change

- Back end new feature.
- Bug fix.
- Refactoring.

## Changes

- New class and object for DashAI registry: `ComponentRegistry`. 
  - The new registry is composed of a two-level dictionary, where the external level is a dictionary of types (Task, Model DataLoader, etc...) whose values contain the registered components associated to each type.
  - The dictionary supports queries to select specific types of components, to ignore others, to select components related to others, and to select components that inherit a specific class.
  - The registry contains an unique internal relationship manager. This object is in charge of managing and querying the relationships between the various components. This object is implemented in the RelationshipManager class as a double dictionary. The first dictionary is a mapping between a component and all others related to it, while the second represents the inverse relationships to it.
- Now all components (configurable or not) are represented as a component dict object. The component dict is a dictionary that unifies all possible registerable objects in DashAI and contains (hopefully) all the attributes of each component: the name, its type, its description, its representative class, whether it is configurable or not through a flag, its schema if applicable, and if so, its type. For example: 
```python
    {
        "name": "Component1",         # Component name.
        "type": "ComponentType",      # Component type.
        "class": Component1,          # Component class.
        "configurable_object": False, # True if the object is a configurable one.
        "schema": {...},              # Configurable object schema if applies.
        "description": "...",         # An object description.
    }
```
- All component-related APIs were updated. All of them changed their responses to component dict or a list of them. 
The getter of several components changed some of its parameters. All the responses of this method are lists of component dict.
Note that the component dict is serialized without the class, since there are no clear rules for transforming a class into json.

Given some limitations regarding the registration of generic components, now:

- Each DashAI component base class is required to implement a `TYPE`.
- Configurable components must implement the class method `get_schema`, which must return the configuration schema. All objects that do not implement get_schema will be considered as non-configurable.
- Now the name of the component is derived directly from the name of the classes. `NAME` or `name` class attribute is not necessary.
- Given the generalization of relations between components, the relations must now be represented through a list of relations declared as a class attribute in a variable `COMPATIBLE_COMPONENTS`. For instance, for TabularClassificationTask, each model should contain an attribute class `COMPATIBLE_COMPONENTS = ["TabularClassificationTask"]` (which is extended from `TabularClassificationModel` class).

## How to Test

- To test the new registry, run ` pytest tests/back/registries/test_registry.py `.
- To test the relationship manager, run `pytest tests/back/registries/test_relationship_manager.py`.
- To test the component API, run `pytest tests/back/api/test_components_api.py `.

## Notes

- There are many decisions implemented. If you find any that you don't agree with, let's discuss them.
- It is probably a very long PR 😥, if you have any questions please contact me directly.
- Probably the description of this PR can be used as a guide for the developer, so it would be a good idea to keep an eye on it.


DashAI/back/api/api_v1/endpoints/components.py
```diff
@@ -1,37 +1,66 @@
...
+    return {key: value for key, value in component_dict.items() if key != "class"}
+
+
 @router.get("/")
 async def get_components(
-    component_type: str | None = None,
-    task_name: str | None = None,
+    select_type: str | None = None,
+    ignore_type: str | None = None,
+    related_component: str | None = None,
```
cristian-tamblay: What do you think about these being enums? I don't know if it has any benefits, but in my head it makes some sense

DashAI/back/api/api_v1/endpoints/components.py
```diff
@@ -46,71 +75,74 @@ async def get_components(
...
+
+    return [
+        _delete_class(component_dict) for component_dict in selected_components.values()
+    ]
 
 
 @router.get("/{id}/")
 def get_component_by_id(id: str) -> dict:
-    """Return a component using its id.
+    """Return an specific component using its id.
```
cristian-tamblay: Who assigns ids to components?

cristian-tamblay:
In general this PR looks really good, i hope this is our last iteration over the registries

DashAI/back/api/api_v1/endpoints/components.py
pbadillatorrealba: Enum = list of types?

If so, in fact the back function accepts lists of types. 

For simplicity I left it as only one parameter, but as this is hopefully the last iteration of this feature, then maybe I'll leave it ready right away.
I will add this new functionality and let you know.

DashAI/back/api/api_v1/endpoints/components.py
pbadillatorrealba: The id is the name of the class (obtained as `some_class.__name__`).

I will add that definition in the endpoint docstring.

DashAI/back/tasks/base_task.py
```diff
@@ -35,7 +35,7 @@ class BaseTask(metaclass=TaskMetaClass):
 
     # task name, present in the compatible models
     name: str = ""
```
inunezn7: This should be deleted, right?

inunezn7: if so, it will be also necessary to change self.name for "self._ _ name _ _" in lines 71 and 78

inunezn7: This could mean that a test about the task returning a non-empty schema is missing

DashAI/back/api/api_v1/endpoints/components.py
```diff
@@ -1,37 +1,66 @@
...
-    """Retrieves components from the register.
+    """Retrieve components from the register according to the provided parameters.
+
+    When all parameters are None, the method return all registered components.
+    If more than one parameter is specified, then the method returns the intersection
+    between the retrived components.
 
     The components returned will depend on the parameters of the request.
-    If all parameters are None, then the method will return all registered components.
+    If all parameters are None, then the method returns all registered components.
```
inunezn7, line 46: Same as line 41

DashAI/back/api/api_v1/endpoints/components.py
```diff
@@ -46,71 +75,74 @@ async def get_components(
...
-        }
+    # 2. ignore the requested types
+    if ignore_type is not None:
+        selected_components = _intersect_component_lists(
+            selected_components,
+            component_registry.get_components_by_type(ignore=ignore_type),
+        )
 
-    return [component.get_schema() for component in selected_components.values()]
+    # 3. select only the task related components
```
inunezn7:
```suggestion
    # 3. select only the components related to "related_component"
```

DashAI/back/registries/component_registry.py
```diff
@@ -0,0 +1,401 @@
...
+        base_classes_cantidates = [
+            ancestor_cls
+            for ancestor_cls in component_base_ancestors
+            if hasattr(ancestor_cls, "TYPE")
+        ]
+
+        # check if there is only one type candidate.
+        if len(base_classes_cantidates) == 0:
+            raise TypeError(
+                f"Component {new_component.__name__} does not a DashAI base class with "
```
inunezn7:
```suggestion
                f"Component {new_component.__name__} does not inherit from any DashAI base class with "
```

DashAI/back/registries/component_registry.py
```diff
@@ -0,0 +1,401 @@
...
+        new_component : Type
+            The object to be registred.
+
+        Raises
+        ------
+        TypeError
+            If the provided component is not a class.
+        TypeError
+            If some task that the component declares compatible does not exist in the
+            taskm registry.
```
inunezn7:
```suggestion
            task registry.
```

DashAI/back/registries/component_registry.py
```diff
@@ -0,0 +1,401 @@
...
+            "description": getattr(new_component, "DESCRIPTION", None),
+        }
+
+        if base_type not in self._registry:
+            self._registry[base_type] = {new_component.__name__: new_register_component}
+        else:
+            self._registry[base_type][new_component.__name__] = new_register_component
+
+        if hasattr(new_component, "COMPATIBLE_COMPONENTS"):
+            for compatible_task in new_component.COMPATIBLE_COMPONENTS:
```
inunezn7:
```suggestion
            for compatible_component in new_component.COMPATIBLE_COMPONENTS:
```

DashAI/back/registries/component_registry.py
```diff
@@ -0,0 +1,401 @@
...
+                self._relationship_manager.add_relationship(
+                    new_component.__name__,
+                    compatible_task,
```
inunezn7:
```suggestion
                    compatible_component,
```

DashAI/back/api/api_v1/endpoints/components.py
```diff
@@ -1,37 +1,66 @@
...
 
     Parameters
     ----------
-    component_type : str | None, optional
-        If specified, the function will return only the components belonging to that
-        component type (e.g., task, model, dataloader, etc...), by default None.
-    task_name : str | None, optional
-        If specified, the function will return only the components related with
-        the task (e.g., TabularClassification, Translation), by default None.
+    select_types: Annotated[list[str] | None, Query()], optional
```
inunezn7, line 50: Just to know, why did you change this name? It sounded more natural to me as component_types

inunezn7 (approved):
The registry looks great. The generalization also makes the code more understandable. 
I just fixed some typos and noted a couple of lines that need to be updated to the new convention about class names.
While reviewing the code I wondered: Could be possible to put all the schemas in just one folder? Since components have to have unique names, I think it is possible. In that way, the method for all configurable objects could be the same and we don't have to define it in every class.

inunezn7 (approved):
The registry looks great. The generalization also makes the code more understandable. 
I just fixed some typos and noted a couple of lines that need to be updated to the new convention about class names.
While reviewing the code I wondered: Could be possible to put all the schemas in just one folder? Since components have to have unique names, I think it is possible. In that way, the method for all configurable objects could be the same and we don't have to define it in every class.

cristian-tamblay (changes requested):
Tests are failing after we changed the input type to a list. I think this is the last change needed

DashAI/back/api/api_v1/endpoints/components.py
pbadillatorrealba: Ok

DashAI/back/registries/component_registry.py
pbadillatorrealba: Ok!

DashAI/back/api/api_v1/endpoints/components.py
pbadillatorrealba: I changed it because I also added the option to ignore types.
So, the select was more explicit to say that I want to select only certain types.

pbadillatorrealba:
@inunezn7 , yes, it is possible that we could order the schemes better, but I think it would be prudent to look at it in another pr.

I have already resolved all your comments :)

DashAI/back/tasks/base_task.py
pbadillatorrealba: Changes applied.

pbadillatorrealba:
@cristian-tamblay , all tests were fixed! :)
The problem was caused by an outdated version of fastapi and some of its dependencies.

The branch is ready to be merged.

cristian-tamblay (approved):

PR #81, opened 2023-06-02
Remove craco

maximi4321:
# Summary

Removes the usage of [craco](https://www.npmjs.com/package/@craco/craco)

## Type of change

- Refactoring.

## Changes

- Uninstall craco
- Delete `DashAI/front/craco.config.js` file
- Modify `package.json` so it uses `react-scripts` instead of `craco`

## How to Test

Test `yarn start`
1. Go to `DashAI/DashAI/front` and run `yarn json-server`
2. in another terminal in the same folder of step 1 run `yarn start`
3. The app should work normally.

Test `yarn build`
3. Enter localhost:8000
4. The app should work normally.

Test typescript type checking
1. Follow the test `yarn start` steps.
2. open the file `DashAI/front/src/api/datasets.ts`
3. modify the return type of `getDatasets` function like this:
```
export const getDatasets = async (): Promise<object> => {
  const response = await api.get<IDataset[]>(datasetEndpoint);
  return response.data;
};
```
4. You should see a type checking error in your terminal or browser

## Notes

I deleted the file `craco.config.js`, which enabled type checking in TypeScript.

```
...
typescript: {
      enableTypeChecking: true /* (default value) */,
    },
...
```
I believe that the configuration for type checking is actually handled by `DashAI/front/tsconfig.json` instead. When I deleted that file, the type checking was disabled, but deleting `craco.config.js` had no effect.


pbadillatorrealba (approved):
Looks good.

PR #83, opened 2023-06-02
Small refactoring of tests

Oziel14:
# Summary

We made a refactor to the tests such as renaming, reorganization of some things according to past comments and elimination of resources.

## Type of change

- Refactoring.

## How to Test

It is enough to launch the tests

## Notes

Depends on the PR #78


tests/back/dataloaders/test_dataloaders.py
```diff
@@ -34,65 +35,57 @@ def test_json_dataloader_to_dataset():
...
-        test_dataset_path = "tests/back/dataloaders/wrong_iris.csv"
-        dataloader_test = CSVDataLoader()
-        params = {"separator": ","}
-        with open(test_dataset_path, "r") as file:
-            csv_data = file.read()
-        csv_binary = io.BytesIO(bytes(csv_data, encoding="utf8"))
-        file = UploadFile(csv_binary)
         dataloader_test.load_data("tests/back/dataloaders", params, file=file)
-    assert True
+    os.remove(
```
Riul1999: I think this remove should be inside the load_data function, in a finally statement

Riul1999 (changes requested):
Overall, the changes are positive

But i think there are still some changes to be made:
- The test for dataloaders still create and do not delete the dashaidataset folder, which should be deleted in the last test that used it or in a conftest fixture teardown (after all tests of the dataloaders package run).
- The test for models should create and manage its own dashaidataset, instead of using the one created by the test for dataloaders.

I also leave a comment for the load_data function that I think it is good to think about it.

pbadillatorrealba:
Also remember to resolve conflicts!

Oziel14:
Done!

Riul1999 (approved):

PR #87, opened 2023-06-06
Fix models get_schema

Riul1999:
# Summary

Refactor classmethod _get_schema_ in **Models**. Now its defined in **BaseModel** and returns the schema.

## Type of change

- Refactoring.

## Changes

- Changed the _get_schema_ method for **Models**, it now returns the json params instead of storing it in the _SCHEMA_ field of the class. This follows the expected behavior declared for the **ComponentRegistry** to retrieve the schema of a component.
- Move the _get_schema_ method of the concrete classes to the **BaseModel** class, it uses the _\_\_name\_\__ property of the class to get the correct schema..
- Add brief description (from scikit-learn) about **RandomForestClassifier**.
- Add test _test_get_schema_from_model_ inside _test_tabular_class_models.py_ to test _get_schema_ beahavior.

## How to Test

To test this you need to type in the terminal 

`pytest .\tests\back\models\test_tabular_class_models.py::test_get_schema_from_model `


pbadillatorrealba (approved):
Nice fix and very brief to review!

Keep an eye on the future: when scikit is removed from the dash core, this whole set of tests will probably fail.

cristian-tamblay (approved):
:ok_hand: PR

PR #91, opened 2023-06-09
Asynchronous Runner

Riul1999:
# Summary

Refactor the **Runner** to run _asynchronously_ in the backend, this allows requests to be made while a **Run** is being processed.

## Type of change

- Back end new feature.
- Refactoring.

## Changes

- Created the _execute_runner_ function inside _core/runner.py_, this function is in charge of executing a **Run** and storing all data of interest.
- Refactored the runner endpoint, now it just checks if all the **Components** and database objects are available and then passes it to _execute_run_.
- In the endpoint runner, the execution of _execute_run_ is performed asynchronously, using FastAPI **BackgroundTasks**.
- New RunStatus **DELIVERED** has been added, this status is used when the execution has been passed to the background (after a request from the runner endpoint) but the execution is not processed by the _execute_run_ function.
- New **Run** field _delivery_time_  to keep track of when the user sent a **Run** to be executed.
- Modified _test_runner_ to check the new **RunStatus** and times. Now it only checks if the times are not None instead of different because the last approach sometimes fails.

## How to Test

To test it's necesary to run `pytest ./tests/back/api/test_runner.py`


DashAI/back/database/models.py
```diff
@@ -81,9 +81,15 @@ class Run(Base):
     status: Mapped[Enum] = mapped_column(
         Enum(RunStatus), nullable=False, default=RunStatus.NOT_STARTED
     )
+    delivery_time: Mapped[DateTime] = mapped_column(DateTime, nullable=True)
     start_time: Mapped[DateTime] = mapped_column(DateTime, nullable=True)
     end_time: Mapped[DateTime] = mapped_column(DateTime, nullable=True)
 
+    def run_delivery(self):
```
pbadillatorrealba: Why not to call the state changer method with a slightly more descriptive name as `set_status_as_delivered`?
This change could be easily extended to the remaining state changing methods. (`set_status_as_started` and `set_status_as_finished`).

```suggestion
    def set_status_as_delivered(self):
```

tests/back/api/test_runner.py
```diff
@@ -144,12 +144,14 @@ def test_exec_runs(client: TestClient, run_id: int):
 
     response = client.get(f"/api/v1/run/{run_id}")
     data = response.json()
-    assert data["train_metrics"] is not None
-    assert data["validation_metrics"] is not None
-    assert data["test_metrics"] is not None
+    assert type(data["train_metrics"]) is dict
+    assert type(data["validation_metrics"]) is dict
+    assert type(data["test_metrics"]) is dict
```
pbadillatorrealba: Use `isinstance` instead:

```suggestion
    assert isinstance(data["train_metrics"], dict)
    assert isinstance(data["validation_metrics"], dict)
    assert isinstance(data["test_metrics"], dict)
```

DashAI/back/api/api_v1/endpoints/runner.py
```diff
@@ -51,51 +53,35 @@ async def execute_run(run_id: int, db: Session = Depends(get_db)):
...
-    run_instance.save(run_path)
-    run.run_path = run_path
+    ).values()
+
+    # Mark run as delivered
+    run.run_delivery()
     db.commit()
 
+    # Execute the run
+    background_tasks.add_task(
```
pbadillatorrealba, line 77: What happens if an exception is raised inside the background task?
Where do you set a run as failed?
My guess is that it is missing a try-except somewhere. 
This try-except should log some error message and modify the run status as failed.

pbadillatorrealba (changes requested):
The implementation looks good and is easy to read :)

I left some code style and names comments and also a question about the error handling.

DashAI/back/core/runner.py
```diff
@@ -0,0 +1,70 @@
...
+        model.save(run_path)
+        run.run_path = run_path
+        db.commit()
+    except Exception as e:
+        log.exception(e)
+        run.set_status_as_error()
+        db.commit()
+        raise RunnerError(
+            "Internal runner error",
+        ) from e
```
cristian-tamblay: I think this exception catching is too general, and the raise RunnerError makes it even worse, now the user wont know what caused the exception. A bare raise is better than raise RunnerError(...). 

I know this is a hard to solve problem, we can at least do the following: we could separate different exceptions such as failed in training or metric evaluation. A good aproach for this is making the try: catch: blocks have less code, as an example, just
```
try:
    model.fit()
except Exception as e:
    log.exception(e)
    run.set_status_as_error()
    db.commit()
    raise RunnerError("Training failed")
finally:...
...
try:
    model.metrics(...)
except Exception as e:
    log.exception(e)
    run.set_status_as_error()
    db.commit()
    raise RunnerError("Metric calculation failed")
finally:...
```

tests/back/api/test_runner.py
```diff
@@ -39,12 +40,35 @@ def load(self, filename):
...
+        return
+
+    def format_data(self, data):
+        return data
+
     def predict(self, x):
         return {}
 
+    def fit(self, x, y):
+        raise Exception("Allways fails")
```
cristian-tamblay: typo
```suggestion
        raise Exception("Always fails")
```

cristian-tamblay (changes requested):
PR is almost good to be merged, I think we should think a bit how to handle exceptions in training. Give it a though and we merge.

cristian-tamblay (approved):
Exception handling is perfectable, but its good enough for this iteration

PR #92, opened 2023-06-09
Save dataset sample in dataset table

nataliabarca:
# Summary

Added a new column called feature_names to the dataset table to save feature names. The column is nullable and saves data in JSON format.

The purpose of this change is to to be able to easily retrieve these values  when using the Prediction tab. The idea is to display a table of the feature names and fields that the user can populate with values to make a prediction.

## Changes

- Added `feature_names` column in Dataset model.
- Modified endpoint `/api/v1/dataset/` to save feature names in dataset table.
- Modified `test_create_csv_dataset` to check that data is saved correctly.

## How to Test

Run `pytest ./tests/back/api/test_datasets.py`


pbadillatorrealba:
Hi @nataliabrc ,

Could you provide more context about your contribution?

I also some questions:
- Sample is just one example? Or is a list/array/etc... of examples? If so, how many examples will you keep? Will you choose them randomly? Will they be the first/last n? 
- Why not have a dataset method called sample, which dinamically opens the dataset and send a generated sample to the front?. It could also mimic the pandas api and send the head or tail rows.

nataliabarca:
Hi @pbadillatorrealba,
I have updated the summary of this pull request to give it more context.

Regarding your questions, I have modified this feature to only save feature names in the dataset table instead of one sample. As you said, it is better to create a dataset method to generate samples dinamically 👌

pbadillatorrealba:
Nice, I see it clear now :)
One last thing: You mentioned you could create a dataset with null feature names. 
Could you add a test that checks this functionality?

nataliabarca:
Hi,
I have modified the new column (features_name) so that it is not null and created a test where a dataset is created with a csv without header names.
I decided to make the column not null because the endpoint for uploading datasets always generates an input_columns list (even if it's empty), but if you have any feedback on this solution or the new test, please let me know :)

pbadillatorrealba (approved):

cristian-tamblay:
This is a nice contribution and its ready to be merged, but i'll wait until Oziel's PRs are accepted so we don't have compatibility issues with the new database (we don't have a makemigrations yet).

PR #93, opened 2023-06-10
Text classification task and model

Oziel14:
# Summary

The text classification task was incorporated with its respective validator and prepare for task. For this purpose, tests were performed. On the other hand, a transformer type model was incorporated with its respective fit, predict, save and load methods. For the latter, several tests were also performed.

Requires an update of the libraries as it uses a new call accelerate

## Type of change
- Back end new feature.

## How to Test
In test_tasks there are the new tests for the text classification task, while in the file test_text_class_models.py there are the tests corresponding to the model.

## Notes
- This pr depends on the #83 
- An important change, but that will be incorporated soon, is that the fit and predict method, as this model does, will only receive the dataset, and inside it will do the format_data and what is necessary.
- Tests may take longer as the model is fine-tuned, although little data was used to make this process more efficient.


DashAI/back/models/hugging_face/distilbert_transformer.py
```diff
@@ -0,0 +1,176 @@
...
+        see if it was fine-tuned.
+        """
+        self.model_name = "distilbert-base-uncased"
+        self.tokenizer = DistilBertTokenizer.from_pretrained(self.model_name)
+        self.model = (
+            model
+            if model is not None
+            else DistilBertForSequenceClassification.from_pretrained(self.model_name)
+        )
+        self.fitted = bool(model is not None)
```
pbadillatorrealba: Why you has to cast the result of that expression to bool?
According of what I know, ´in´always returns a bool.

DashAI/back/models/hugging_face/distilbert_transformer.py
```diff
@@ -0,0 +1,176 @@
...
+        )
+        train_dataset.set_format(
+            "torch", columns=["input_ids", "attention_mask", "labels"]
+        )
+
+        # Arguments for fine-tuning
+        training_args = TrainingArguments(
+            output_dir="DashAI/back/user_models/temp_checkpoints_distilbert",
+            save_steps=1,
+            save_total_limit=1,
```
pbadillatorrealba, line 104: Shouldn't those arguments be configurable hyperparameters?

DashAI/back/models/parameters/models_schemas/DistilBERT.json
```diff
@@ -0,0 +1,63 @@
+{
+    "additionalProperties": false,
+    "error_msg": "DistilBert parameters must be one(s) of ['epochs', 'batch size', 'weight decay', 'learning_rate', 'device'].",
+    "description": "Distilbert is a transformer that allows you to classify text in English.",
```
pbadillatorrealba: Include in the description that this model is uncased and its based on huggingface distilbert-base-uncased model.

DashAI/back/models/hugging_face/distilbert_transformer.py
```diff
@@ -0,0 +1,176 @@
...
+        """
+
+        train_dataset = dataset["train"]
+        input_column = train_dataset.inputs_columns[0]
+        output_column = train_dataset.outputs_columns[0]
+
+        tokenizer_func = self.get_tokenizer(input_column, output_column)
+        train_dataset = train_dataset.map(
+            tokenizer_func, batched=True, batch_size=len(train_dataset)
+        )
```
pbadillatorrealba: Why the batch size has the size of the train dataset?
What happens if the train dataset is too big?

DashAI/back/models/hugging_face/distilbert_transformer.py
```diff
@@ -0,0 +1,176 @@
...
+            save_total_limit=1,
+            **self.training_args,
+        )
+
+        # The Trainer class is used for fine-tuning the model.
+        trainer = Trainer(
+            model=self.model,
+            args=training_args,
+            train_dataset=train_dataset,
+        )
```
pbadillatorrealba, line 113: Wouldn't it be good in this case to use the validation dataset as well?
Currently, how do you are evaluating overfitting?
Also, wouldn't it be good to add early stopping?

DashAI/back/models/hugging_face/distilbert_transformer.py
```diff
@@ -0,0 +1,176 @@
...
+
+        trainer.train()
+        self.fitted = True
+        shutil.rmtree(
+            "DashAI/back/user_models/temp_checkpoints_distilbert", ignore_errors=True
+        )
+        return
```
pbadillatorrealba: It is not necessary to specify a return if your function does not return anything.

DashAI/back/models/hugging_face/distilbert_transformer.py
pbadillatorrealba, line 113: Wouldn't it be good in this case to use the validation dataset as well?
Currently, how do you are evaluating overfitting?
Also, wouldn't it be good to add early stopping?

DashAI/back/models/hugging_face/distilbert_transformer.py
```diff
@@ -0,0 +1,176 @@
...
+            probabilities.extend(probs.detach().cpu().numpy())
+        return np.array(probabilities)
+
+    def save(self, filename=None):
+        self.model.save_pretrained(filename)
+
+    @classmethod
+    def load(cls, filename):
+        model = DistilBertForSequenceClassification.from_pretrained(filename)
+        return cls(model=model)
```
pbadillatorrealba, line 173: This method are deprecated and they are not required anymore.
```suggestion
```

DashAI/back/tasks/text_classification_task.py
```diff
@@ -1,14 +1,21 @@
...
 
 from DashAI.back.tasks.base_task import BaseTask
 
 
 class TextClassificationTask(BaseTask):
-    """Base class for text classifitacion task."""
+    """
+    Class to represent the Text Classification task.
+    Here you can change the methods provided by class Task.
+    """
```
pbadillatorrealba:
```suggestion
    """Base class for Text Classification Task
    """
```

DashAI/back/tasks/text_classification_task.py
```diff
@@ -1,14 +1,21 @@
...
+
+    name: str = "TextClassificationTask"
```
pbadillatorrealba: The name is now inferred from the class name, so this variable is not needed.

```suggestion
```

DashAI/back/tasks/text_classification_task.py
pbadillatorrealba, line 1: Could you also remove the create method in this pr?

DashAI/back/tasks/text_classification_task.py
```diff
@@ -22,103 +29,22 @@ def create():
...
+        """Change the column types to suit the tabular classification task.
+        A copy of the dataset is created.
 
-        Args:
-            dataset (DatasetDict): Uploaded dataset in a DatasetDict.
-            class_column (str/int): Name or index of class column of the dataset.
+        Parameters
+        ----------
+        dataset : DatasetDict
+            Dataset to be changed
```
pbadillatorrealba: Maybe , `Dataset to be configured?`

tests/back/tasks/test_tasks.py
```diff
@@ -98,3 +101,43 @@ def test_not_prepared_task():
...
+        datasetdict, 0.7, 0.1, 0.2, class_column=outputs_columns[0]
+    )
+    return separate_datasetdict
+
+
+def test_create_text_task():
+    try:
+        TextClassificationTask.create()
+    except Exception as e:
+        pytest.fail(f"Unexpected error in test_create_tabular_task: {repr(e)}")
```
pbadillatorrealba: This test is not needed.

DashAI/back/models/hugging_face/distilbert_transformer.py
```diff
@@ -0,0 +1,176 @@
...
+class DistilBertTransformer(BaseModel, TextClassificationModel):
+    """
+    Pre-trained transformer DistilBERT allowing English text classification
+    """
+
+    @classmethod
+    def get_schema(cls):
+        with open("DashAI/back/models/parameters/models_schemas/DistilBERT.json") as f:
+            cls.SCHEMA = json.load(f)
+        return cls.SCHEMA
```
pbadillatorrealba, line 27: It is not necessary to return the schema once it is assigned into the class attributes.
```suggestion
```

pbadillatorrealba (changes requested):
I left some comments about code style and deprecated functionalities.
Also, I'm in doubt with the fit method and the use of the validation set.
For that, I put some comments in the model trainer that I think we should discuss in some meeting.

DashAI/back/models/hugging_face/distilbert_transformer.py
Oziel14: For some reason precommit ruff made me leave it in that format, so I changed it.

DashAI/back/models/hugging_face/distilbert_transformer.py
Oziel14: I left it with those hyperparameters because for now, according to what we discussed with Felipe, we are only going to worry about the final state of the model and not the checkpoints, so this way we don't have several checkpoint folders lying around.

DashAI/back/models/hugging_face/distilbert_transformer.py
Oziel14: As in the hyperparameters issue, according to Felipe's comments, everything related to intermediate training states of the model will be ignored for now, such as early stopping.

DashAI/back/models/hugging_face/distilbert_transformer.py
Oziel14: About this method I think it's a bit counter-intuitive to have a method called get and it doesn't deliver anything, we could talk about it. I think it would be better to just have that attribute without the need to call get.

DashAI/back/models/hugging_face/distilbert_transformer.py
Oziel14: My problem with eliminating this method is that you would not be able to test as well that the save method is working.

DashAI/back/models/hugging_face/distilbert_transformer.py
pbadillatorrealba: Ok, let's do it in a later pr.

pbadillatorrealba (approved):

tests/back/models/ImdbSentimentDatasetSmall.json
cristian-tamblay, line 1: We need to do something with these datasets files to test our features, the same datasets are even getting duplicated at different pytest modules.
@pbadillatorrealba suggested that we use another repo to upload all these datasets and then fetch from there, because this will not scale and it will make our installation really slow.
I will approve but we need to have this in mind

cristian-tamblay (approved):

PR #94, opened 2023-06-10
Update fit and predict input

Oziel14:
# Summary
The fit function now receives a DashAIDataset and internally separates and formats it.  The predict part also receives a DashAIDaset. Now to calculate the metrics you just need to pass the DashAIDataset of the partition you want together with the predictions. The corresponding changes were made to the tests

## Type of change

- Back end new feature.
- Refactoring.

## How to Test

In the corresponding files the tests of the tabular models and the metrics

## Notes
This PR depends on #93


pbadillatorrealba:
I will review deeply this pr when #93 is approved, but I also take this opportunity to ask: In which place we can include the validation dataset when the trainer require it (for example, in gradient based models)?

Oziel14:
In the first instance I had in mind to receive the complete DatasetDict instead of a DashAIDataset alone, this way we would have full access to all the partitions and could use the validation set. In that case we would have to change a little the logic, and for example at API level only call predict once to obtain the metrics in all the sets.

DashAI/back/metrics/classification_metric.py
```diff
@@ -9,7 +13,7 @@ class ClassificationMetric(BaseMetric):
     COMPATIBLE_COMPONENTS = ["TabularClassificationTask"]
 
 
-def validate_inputs(true_labels: list, pred_labels: list):
+def validate_inputs(true_labels: ndarray, pred_labels: list):
```
pbadillatorrealba: Use `np.ndarray` instead importing just `ndarray`

DashAI/back/metrics/classification_metric.py
```diff
@@ -21,3 +25,25 @@ def validate_inputs(true_labels: list, pred_labels: list):
     """
     if len(true_labels) != len(pred_labels):
         raise ValueError("The length of the true and predicted labels must be equal.")
+
+
+def prepare_to_metric(true_labels: DashAIDataset, probs_pred_labels: list):
```
pbadillatorrealba: `probs_pred_labels` should not be an np.array?
If so, you should change the description to something like "A two-dimensional matrix in which each column represents a class and the row values represent the probability that an example belongs to the class associated with the column

Also, I think that is good idea to change `true_labels` to `dataset`, since that is what you are finally providing to the method. If so, you should also update the docstring definition (something like "A dashai dataset with labels included".


```suggestion
def prepare_to_metric(dataset: DashAIDataset, probs_pred_labels: np.ndarray) -> Tuple[np.array, np.array]:
```

DashAI/back/metrics/classification_metric.py
```diff
@@ -21,3 +25,25 @@ def validate_inputs(true_labels: list, pred_labels: list):
...
+    """Format labels to be used in metrics
```
pbadillatorrealba:
```suggestion
    """Prepare true and prediced labels to be used later in metrics.
```

DashAI/back/metrics/classification_metric.py
```diff
@@ -21,3 +25,25 @@ def validate_inputs(true_labels: list, pred_labels: list):
...
+    Parameters
+    ----------
+    true_labels : DashAIDataset
+        True labels
+    probs_pred_labels : list
+        Probabilities of belonging to the class according to the model
+
+    Returns
+    -------
+    tuple
```
pbadillatorrealba:
```suggestion
    Tuple[np.ndarray, np.ndarray]
```

DashAI/back/metrics/classification_metric.py
```diff
@@ -21,3 +25,25 @@ def validate_inputs(true_labels: list, pred_labels: list):
...
+        A tuple with the true and predicted labels in numpy format
```
pbadillatorrealba:
```suggestion
        A tuple with the true and predicted labels in numpy format.
```

DashAI/back/metrics/classification/accuracy.py
```diff
@@ -13,12 +13,12 @@ class Accuracy(ClassificationMetric):
...
     @staticmethod
-    def score(true_labels: list, probs_pred_labels: list):
+    def score(true_labels: DashAIDataset, probs_pred_labels: list):
         """Calculates the accuracy between true labels and predicted labels
 
         Parameters
         ----------
-        true_labels : list
+        true_labels : Dataset
             True labels
```
pbadillatorrealba:
```suggestion
        true_labels : DashAIDataset
            A DashAI dataset with labels.
```

DashAI/back/metrics/classification/accuracy.py
```diff
@@ -13,12 +13,12 @@ class Accuracy(ClassificationMetric):
     """
 
     @staticmethod
-    def score(true_labels: list, probs_pred_labels: list):
+    def score(true_labels: DashAIDataset, probs_pred_labels: list):
```
pbadillatorrealba: Check `prepare_to_metric` comments and apply here too.

```suggestion
    def score(dataset: DashAIDataset, probs_pred_labels: np.ndarray) -> float:
```

DashAI/back/metrics/classification_metric.py
```diff
@@ -1,3 +1,7 @@
+import numpy as np
+from numpy import ndarray
```
pbadillatorrealba:
```suggestion
```

DashAI/back/models/scikit_learn/sklearn_like_model.py
```diff
@@ -19,35 +19,29 @@ def load(filename):
...
-            formatted_data_to_model[f"{split}"] = {
-                "input": input_data,
-                "output": output_data,
-            }
-
-        return formatted_data_to_model
-
-    def fit(self, x, y):
+        data_in_pandas = dataset.to_pandas()
+        x = data_in_pandas.loc[:, ~data_in_pandas.columns.isin(dataset.outputs_columns)]
```
pbadillatorrealba: Why not to use `dataset.input_columns`?

DashAI/back/core/runner.py
```diff
@@ -69,8 +66,8 @@ def execute_run(
             model_metrics = {
                 split: {
                     metric.__name__: metric.score(
-                        formated_dataset[split]["output"],
-                        model.predict(formated_dataset[split]["input"]),
+                        dataset[split],
+                        model.predict(dataset[split]),
```
pbadillatorrealba: Why not to predict just one time and then use this values to calculate the metric?
Here you are doing the same prediction the number of metrics associated with the task.

pbadillatorrealba: Also, why here you use `dataset` directly and not `prepared_dataset`?

pbadillatorrealba (changes requested):
I think the change to DashAIDataset was quite elegant :)

I left some comments about some operations, names, types and documentation.

Note especially the change of `prepare_to_metric`.
In order not to repeat the same thing, all score methods their signature and types should be updated similar as the indicated in the `prepare_to_metric` method.

pbadillatorrealba (approved):

PR #96, opened 2023-06-12
Use of run endpoint

maximi4321:
# Summary

 
The "runs" endpoint is used to create new runs.

## Type of change

- Front end new feature

## Changes

- The file `api/run.ts` is created to contain the functions to get and create runs
-  The function to create runs is used in `experiments/newExperimentModal.jsx`
- Temporary functions are created, which will be modified once the compatibility issues with the backend endpoints are resolved.

## How to Test

3. Create a new experiment (follow the steps in the video in the "Screenshots" section)
4. To verify that your runs were created go to `localhost:8000/api/v1/docs` and try the "Get Runs" endpoint.

## Screenshots

[run_endpoint.webm](https://github.com/DashAISoftware/DashAI/assets/53657198/a82cc685-7f63-4e76-b90f-8d2225110672)

## Notes

In order to make the application functional, some modifications have been made to allow POST request functions to send certain parameters in the URL:
- `createExperimentTemp` in `api/experiment.ts`
- `createRun` in `api/run.ts`

These modifications will be corrected once the compatibility issues with the backend endpoints are resolved.


DashAI/front/src/components/experiments/NewExperimentModal.jsx
```diff
@@ -51,15 +52,47 @@ export default function NewExperimentModal({ open, setOpen }) {
...
+  const uploadRuns = async (experimentId) => {
+    for (const run of newExp.runs) {
+      try {
+        await createRunRequest(
+          experimentId,
+          run.type,
+          run.nickname,
+          run.params,
+          "",
+        );
```
pbadillatorrealba, line 64: I think that this type of operation can be dangerous because it can saturate the server.

Wouldn't it be better for the back to have an endpoint where you send it all the runs in a batch and it would take care of gluing them together?

DashAI/front/src/api/run.ts
```diff
@@ -0,0 +1,23 @@
+import api from "./api";
+import type { IRun } from "../types/run";
+
+export const getRuns = async (experimentId: string = ""): Promise<IRun[]> => {
+  const params = experimentId !== "" ? { experiment_id: experimentId } : {};
```
pbadillatorrealba, line 5: Is it possible that experimentID is an empty string?

In other words, is it possible for any run not to have an associated experiment?

DashAI/front/src/api/experiment.ts
```diff
@@ -17,6 +17,18 @@ export const createExperiment = async (formData: FormData): Promise<object> => {
...
 
+// temporary function, will be deleted once the compatibility issues with the backend are resolved.
+export const createExperimentTemp = async (
+  datasetId: number,
+  taskName: string,
+  name: string,
+): Promise<IExperiment> => {
+  const response = await api.post<IExperiment>(
+    `${endpointURL}/?dataset_id=${datasetId}&task_name=${taskName}&name=${name}`,
+  );
```
pbadillatorrealba, line 28: Remember that in a post you should send the data in the body, not as query params :)

I also ask, it is ok the endpoint? I didn't see any valid path to the server.

pbadillatorrealba (commented):
Some comments. 
There is one that is more interesting than the rest:. the runner api design.

PR #103, opened 2023-06-16
Back/update imagedataloader

Oziel14:
# Summary

Some changes were made to the imagedataloader to make it usable, some tests were also added for this dataloader.

## Type of change

- Back end new feature.
- Bug fix.

## How to Test

Just run the tests in the file test_dataloders.py

## Notes

Depends on the PR #94


pbadillatorrealba:
@Oziel14 , please update to the last staging version.

Oziel14:
@pbadillatorrealba Done!

DashAI/back/api/api_v1/endpoints/datasets.py
```diff
@@ -26,7 +27,11 @@
 
 # TODO: Implement Dataloader Registry
 
-dataloaders = {"CSVDataLoader": CSVDataLoader(), "JSONDataLoader": JSONDataLoader()}
+dataloaders = {
+    "CSVDataLoader": CSVDataLoader(),
+    "JSONDataLoader": JSONDataLoader(),
+    "ImageDataloader": ImageDataLoader(),
+}
```
pbadillatorrealba, line 34: Is the registry for dataloaders not working?
How difficult would it be in a future pr to move the dataloaders to the centralized registry?

DashAI/back/dataloaders/classes/dataloader.py
```diff
@@ -141,41 +146,62 @@ def split_dataset(
...
 
-        stratify_column = class_column if stratify else None
+        inputs_columns = dataset["train"].inputs_columns
+        outputs_columns = dataset["train"].outputs_columns
+
+        # Get the number of records
+        n = len(dataset["train"])
+
+        # Generate shuffled indices
+        np.random.seed(seed)
```
pbadillatorrealba, line 156: Why do you fix the seed here if you later fix it in `train_test_split`?

DashAI/back/dataloaders/classes/dataloader.py
```diff
@@ -141,41 +146,62 @@ def split_dataset(
...
-        dataset["test"] = test_valid_split["train"]
-        dataset["validation"] = test_valid_split["test"]
 
-        train_dataset_dict = dataset["train"].to_dict()
-        test_dataset_dict = dataset["test"].to_dict()
-        validation_dataset_dict = dataset["validation"].to_dict()
+        # Convert the indices into boolean masks
+        train_mask = np.isin(np.arange(n), train_indices)
+        test_mask = np.isin(np.arange(n), test_indices)
+        val_mask = np.isin(np.arange(n), val_indices)
```
pbadillatorrealba, line 190: Why not to use `indices` if its the same thing and it was already defined?

```suggestion
        train_mask = np.isin(indices, train_indices)
        test_mask = np.isin(indices, test_indices)
        val_mask = np.isin(indices, val_indices)
```

DashAI/back/dataloaders/classes/dataloader.py
```diff
@@ -141,41 +146,62 @@ def split_dataset(
...
 
-        train_dataset = Dataset.from_dict(train_dataset_dict)
-        test_dataset = Dataset.from_dict(test_dataset_dict)
-        validation_dataset = Dataset.from_dict(validation_dataset_dict)
+        # Create separate tables for each split
+        train_table = table.filter(train_mask)
+        test_table = table.filter(test_mask)
+        val_table = table.filter(val_mask)
 
         separate_dataset_dict = DatasetDict(
```
pbadillatorrealba, line 200: I think this variable could be called `split_dataset_dict` to better represent what it has.

DashAI/back/dataloaders/classes/dataloader.py
```diff
@@ -123,6 +125,9 @@ def split_dataset(
             )
 
         # Value checks
+        if stratify and class_column is None:
+            raise ValueError("Stratify requires that class_column is not none")
```
pbadillatorrealba, line 129:
```suggestion
            raise ValueError("Stratify requires that class_column is not None.")
```

DashAI/back/dataloaders/classes/csv_dataloader.py
```diff
@@ -71,13 +72,14 @@ def load_data(
             dataset = load_dataset("csv", data_files=url, sep=separator)
         elif file:
             files_path = self.extract_files(dataset_path, file)
-            try:
-                if files_path.split("/")[-1] == "files":
+            if files_path.split("/")[-1] == "files":
```
pbadillatorrealba, line 75: What happens if this operation fails?

DashAI/back/dataloaders/classes/image_dataloader.py
```diff
@@ -8,7 +8,7 @@ class ImageDataLoader(BaseDataLoader):
     """Data loader for data from image files."""
 
     def load_data(
-        self, dataset_path: str, file: UploadFile = None, url: str = None
+        self, dataset_path: str, params=None, file: UploadFile = None, url: str = None
```
pbadillatorrealba, line 11: I think the signature and/or documentation of this method needs to be improved.

It is not clear to me, what is dataset_path? 
In that path will the files be stored? If so, then it should be called output_path or something like that.

Also, why specify file and url? 
If the two variables do the same thing only varying the source, then you should unify these parameters in one argument as filepath, as pandas does in its read_* methods (see read_csv).

Also, you added a `params` argument but did not specify what it does it in the documentation nor its possible type.

DashAI/back/dataloaders/classes/json_dataloader.py
```diff
@@ -71,13 +72,14 @@ def load_data(
             dataset = load_dataset("json", data_files=url, field=field)
         elif file:
             files_path = self.extract_files(dataset_path, file)
-            try:
-                if files_path.split("/")[-1] == "files":
+            if files_path.split("/")[-1] == "files":
```
pbadillatorrealba, line 75: This operation apparently repeats itself quite a lot, wouldn't it be better to implement it as a method that is inherited from `BaseDataloader`?

pbadillatorrealba (changes requested):
According to what I saw, this pr had two changes: the change of split to the new DashAIDataset definitions and the implementation of a dataloader for images.
I think the first one looked pretty good but with some details. The second one I think needs to be refined a bit more.
Leave comments for both.

DashAI/back/api/api_v1/endpoints/datasets.py
Oziel14: The truth is that I didn't realize about adding the dataloader to the registry, I think I did it in the next commit.

DashAI/back/dataloaders/classes/csv_dataloader.py
Oziel14: If there is an error in the extract files that will throw the error or do you mean something else?

DashAI/back/dataloaders/classes/image_dataloader.py
Oziel14: change the name to output_path to make it clearer, also remove the params as it is not used in this dataloader. If about the way to upload a dataset I realized that from a url (which in written rigor is the name in the HF Hub) it is done in another way and in fact it is not necessary a specific dataloader. I suggest in that case to remove the URL parameter and then add a specific dataloader for the Hub cases.

DashAI/back/dataloaders/classes/json_dataloader.py
Oziel14: I think it would be better not to generalize it yet in BaseDataloader since only 2 of the 3 have a similar behavior, the imagedataloader is different since it can only come in ZIP format, so it changes the code.

Oziel14:
@pbadillatorrealba Some changes are ready and in others I left comments.

DashAI/back/dataloaders/classes/image_dataloader.py
Oziel14: My mistake about the hub, the way it is used to use the urls, but in the same way to use the hub urls it would be necessary another type of dataloader.

DashAI/back/dataloaders/classes/csv_dataloader.py
pbadillatorrealba: Yes, that thing.
I was telling that the back will raise a strange exception, with no clue of what really failed.
But, I guess that we could fix the error handling in future work :)

DashAI/back/dataloaders/classes/image_dataloader.py
pbadillatorrealba: Yes, I think that we should implement specific dataloaders for huggingface hub datasets. 
I will appoint this in a card and accept this changes.

DashAI/back/dataloaders/classes/json_dataloader.py
pbadillatorrealba: Ok 👍

pbadillatorrealba (approved):

PR #105, opened 2023-06-17
Back/image classification task

Oziel14:
# Summary
It contains the class associated with image classification as well as the appropriate testing. It also contains the incorporation of the Vit model with its correct testing of its fit, predict, load and save methods.

## Type of change

- Back end new feature.
- Documentation.

## How to Test

For testing just look at test_task.py and test_image_class_models.py files.

## Notes
Depends on the PR #103


pbadillatorrealba:
@Oziel14 , please update to the latest staging version and resolve conflicts :)

Oziel14:
@pbadillatorrealba Done!

DashAI/back/models/hugging_face/vit_transformer.py
```diff
@@ -0,0 +1,164 @@
...
+            inputs = self.feature_extractor(
+                images=examples[input_column], return_tensors="pt", size=224
+            )
+            inputs["labels"] = examples[output_column]
+            return inputs
+
+        return preprocess_images
+
+    def fit(self, dataset: DashAIDataset):
+        """Fine-tuning the pre-trained model
```
pbadillatorrealba:
```suggestion
        """Fine-tune the pre-trained model.
```

DashAI/back/models/hugging_face/vit_transformer.py
```diff
@@ -0,0 +1,164 @@
...
+        self.model = (
+            model
+            if model is not None
+            else ViTForImageClassification.from_pretrained(self.model_name)
+        )
+        self.fitted = bool(model is not None)
+        self.training_args = kwargs
+
+    def get_preprocess_images(self, input_column: str, output_column: str):
+        """Preprocess images for model input
```
pbadillatorrealba: Documentation linter claims if you don't put the dot at the end 🫠

```suggestion
        """Preprocess images for model input.
```

DashAI/back/models/hugging_face/vit_transformer.py
```diff
@@ -0,0 +1,164 @@
...
+            args=training_args,
+            train_dataset=dataset,
+        )
+
+        trainer.train()
+        self.fitted = True
+        shutil.rmtree(
+            "DashAI/back/user_models/temp_checkpoints_vit", ignore_errors=True
+        )
+        return
```
pbadillatorrealba: If we follow the scikit-learn convention, this should return to itself (self).

```suggestion
        return self
```

DashAI/back/models/hugging_face/vit_transformer.py
```diff
@@ -0,0 +1,164 @@
...
+    Pre-trained transformer ViT allowing image classification
+    """
+
+    @classmethod
+    def get_schema(cls):
+        with open("DashAI/back/models/parameters/models_schemas/ViT.json") as f:
+            cls.SCHEMA = json.load(f)
+        return cls.SCHEMA
+
+    def __init__(self, model=None, **kwargs):
```
pbadillatorrealba, line 21: If we follow the scikit-learn convention, all parameters received by this model should be explicitly defined. Then, you could group them in training_args dict.

This also gives you the possibility to verify them individually.

```suggestion
    def __init__(self, model=None, epochs: int, batch_size: int, learning_rate: float, device: Literal["gpu", "cpu"], weight_decay: float):
```


I also take this opportunity to ask, in which case model could be used 

Finally, please add a brief description to the parameters in the documentation.

DashAI/back/models/hugging_face/vit_transformer.py
```diff
@@ -0,0 +1,164 @@
...
+        Make a prediction with the fine-tuned model
+
+        Parameters
+        ----------
+        dataset : DashAIDataset
+            DashAIDataset with image data
+
+        Returns
+        -------
+        Numpy Array
```
pbadillatorrealba:
```suggestion
        np.array
```

DashAI/back/models/hugging_face/vit_transformer.py
```diff
@@ -0,0 +1,164 @@
...
+        )
+
+        trainer.train()
+        self.fitted = True
+        shutil.rmtree(
+            "DashAI/back/user_models/temp_checkpoints_vit", ignore_errors=True
+        )
+        return
+
+    def predict(self, dataset: DashAIDataset):
```
pbadillatorrealba:
```suggestion
    def predict(self, dataset: DashAIDataset) -> np.array:
```

DashAI/back/models/hugging_face/vit_transformer.py
```diff
@@ -0,0 +1,164 @@
...
+        """
+        Make a prediction with the fine-tuned model
```
pbadillatorrealba:
```suggestion
        Make a prediction with the fine-tuned model.
```

DashAI/back/models/hugging_face/vit_transformer.py
```diff
@@ -0,0 +1,164 @@
...
+
+        Parameters
+        ----------
+        dataset : DashAIDataset
+            DashAIDataset with image data
```
pbadillatorrealba:
```suggestion
            DashAIDataset with image data.
```

DashAI/back/models/hugging_face/vit_transformer.py
```diff
@@ -0,0 +1,164 @@
...
+
+        probabilities = []
+
+        # Iterate over each batch in the dataset
+        for i in range(len(dataset)):
+            # Prepare a batch of images for the model
+            batch = dataset[i]
+
+            # Make sure that the tensors are in the correct device.
+            batch = {k: v.to(self.model.device) for k, v in batch.items()}
```
pbadillatorrealba, line 142: Just so we think about it with the team later, what would happen if my device fails to move the tensors to the gpu and throws an exception. What would happen to the front end?

DashAI/back/models/parameters/models_schemas/ViT.json
```diff
@@ -0,0 +1,63 @@
...
+                    "minimum": 0,
+                    "default": 5e-5
+                }
+            ]
+        },
+        "device": {
+            "oneOf": [
+                {
+                    "error_msg": "The parameter must be in string format and can be either 'gpu' or 'cpu'.",
+                    "description": "Hardware that will run the training. GPU is recommended for efficiency issues",
```
pbadillatorrealba:
```suggestion
                    "description": "Hardware on which the training is run. If available, GPU is recommended for efficiency reasons. Otherwise, use CPU".,
```

DashAI/back/models/parameters/models_schemas/ViT.json
```diff
@@ -0,0 +1,63 @@
...
+                    "default": "gpu",
+                    "enum": ["gpu", "cpu"]
+                }
+            ]
+        },
+        "weight_decay": {
+            "oneOf": [
+                {
+                    "error_msg": "A number between 0 and 1 must be entered.",
+                    "description": "The weight decay to apply (if not zero) to all layers except all bias and LayerNorm weights in AdamW optimizer.",
```
pbadillatorrealba: It is kind of hard this definition, isn't there a way to make it a little more explanatory?

DashAI/back/tasks/image_classification_task.py
```diff
@@ -0,0 +1,42 @@
...
+    schema: dict = {
+        "inputs_types": [Image],
+        "outputs_types": [ClassLabel],
+        "inputs_cardinality": 1,
+        "outputs_cardinality": 1,
+    }
+
+    @staticmethod
+    def create():
+        return ImageClassificationTask()
```
pbadillatorrealba: This method is supposed to be deprecated.
```suggestion
```

requirements.txt
```diff
@@ -8,9 +8,11 @@ scikit-learn>=1.2.1
...
 evaluate>=0.4.0
 accelerate>=0.20.3
+Pillow>=9.5.0
 
 torch==1.13.0 #+cu116
 --find-links https://download.pytorch.org/whl/torch_stable.html
 transformers>=4.23.1
 sacrebleu>=2.3.1
 sentencepiece>=0.1.97
+Pillow>=9.5.0
```
pbadillatorrealba: I think this dependency is repeated (check line 11)?

```suggestion

```

pbadillatorrealba (changes requested):
Nice implementation!

I left some minor comments about how to implement models according to the scikit-learn convention.

pd: I only checked the model, the dataloader I saw in the previous pr.

DashAI/back/models/hugging_face/vit_transformer.py
Oziel14: In this I have a couple of doubts, because I feel that it is not very extensible if we then add more parameters for the models (because there are many more).  model is used when the model is loaded, this way a new instance of a transformer is initialized but with the new weights.

DashAI/back/models/hugging_face/vit_transformer.py
Oziel14: It's a good question and we should talk about it.

DashAI/back/models/hugging_face/vit_transformer.py
pbadillatorrealba: Ok, we will leave the topic for later.

pbadillatorrealba (approved):

PR #106, opened 2023-06-18
Back/translation task

Oziel14:
# Summary

It contains the class associated with translation as well as the appropriate testing. It also contains the incorporation of the opus-mt-en-es model with its correct testing of its fit, predict, load and save methods. In addition to a small change to the split method of the dataloaders

## Type of change

- Back end new feature.
- Bug fix.

## How to Test

Simply run test_tasks and test_translation_model

## Notes
Depends on the PR #105


DashAI/back/models/hugging_face/opus_mt_en_es_transformer.py
```diff
@@ -0,0 +1,151 @@
...
+
+        # Arguments for fine-tuning
+        training_args = Seq2SeqTrainingArguments(
+            output_dir="DashAI/back/user_models/temp_checkpoints_opus-mt-en-es",
+            save_steps=1,
+            save_total_limit=1,
+            per_device_train_batch_size=self.batch_size,
+            per_device_eval_batch_size=self.batch_size,
+            no_cuda=self.device != "gpu",
+            **self.training_args,
```
pbadillatorrealba: Why not just get `self.batch_size` and `self.device` from `self.training_params` instead of storing them as instance attributes in `__init__()`?

pbadillatorrealba: Or maybe the reverse, instantiate as instance variable the `Seq2SeqTrainingArguments` in the `__init__` as `self.training_args = Seq2SeqTrainingArguments(...)` (similar to how you do it with the tokenizer)?

DashAI/back/models/hugging_face/opus_mt_en_es_transformer.py
```diff
@@ -0,0 +1,151 @@
...
+
+        trainer.train()
+        self.fitted = True
+        shutil.rmtree(
+            "DashAI/back/user_models/temp_checkpoints_opus-mt-en-es", ignore_errors=True
+        )
+        return self
+
+    def predict(self, dataset: DashAIDataset) -> List:
+        """Predicting with the fine-tuned model
```
pbadillatorrealba:
```suggestion
        """Predict using the fine-tuned model.
```

DashAI/back/models/hugging_face/opus_mt_en_es_transformer.py
```diff
@@ -0,0 +1,151 @@
...
+        dataset = dataset.map(encode, batched=True)
+        dataset.set_format(type="torch", columns=["input_ids", "attention_mask"])
+
+        translations = []
+
+        for example in dataset:
+            inputs = {
+                k: v.unsqueeze(0).to(self.model.device) for k, v in example.items()
+            }
+            outputs = self.model.generate(**inputs)
```
pbadillatorrealba, line 137: Perhaps as a side note, doesn't generate also have its own parameters, such as the type of search to be performed (e.g. use beam search)?
I don't know if you have thought about how to deal with this issue?

DashAI/back/models/parameters/models_schemas/OpusMtEnESTransformer.json
```diff
@@ -1,12 +1,12 @@
 {
     "additionalProperties": false,
-    "error_msg": "The Tatoeba Challenge English-Spanish parameters must be one or more of ['epochs', 'batch size', 'weight decay'].",
-    "description": "Tatoeba Challenge English-Spanish is a transformer that allows you to translate sentences from English to Spanish.",
+    "error_msg": "opus-mt-en-es parameters must be one(s) of ['epochs', 'batch size', 'weight decay', 'learning_rate', 'device'].",
+    "description": "opus-mt-en-es is a transformer that allows you to translate texts from english to spanish",
```
pbadillatorrealba:
```suggestion
    "description": "opus-mt-en-es is a transformer pre-trained model that allows translation of texts from English to Spanish.",
```

DashAI/front/src/components/ConfigurableObject/Inputs/ListOfStringsInput.jsx
pbadillatorrealba, line 1: Which of the new configurable objects (of `OpusMtEnESTransformer.json` schema) uses this component?

I don't know if I understood correctly, but it seems that the options are hardcoded in `config`?

DashAI/front/src/components/experiments/ConfigureModelsStep.jsx
pbadillatorrealba, line 1: Note that this fix will probably cause some problems with the current pr (depending on which one you merge first).

DashAI/front/src/components/experiments/EditModelDialog.jsx
pbadillatorrealba, line 1: Same here.

pbadillatorrealba (changes requested):
Overall it looks good. 
I left some comments on some sections that generated doubts and that could be improved.

DashAI/back/models/hugging_face/opus_mt_en_es_transformer.py
Oziel14: To be honest I had ignored it. But in that case it could be requested as an extra argument for the model or it could be a separate configurable object in the future like a tokenizer.

DashAI/front/src/components/ConfigurableObject/Inputs/ListOfStringsInput.jsx
maximi4321: >Which of the new configurable objects (of OpusMtEnESTransformer.json schema) uses this component?

This component is used while rendering the parameter form of the json dataloader in `DashAI/back/dataloaders/params_schemas/jsondataloader.json`.

>I don't know if I understood correctly, but it seems that the options are hardcoded in config?

Yes, the options of output_columns are hardcoded.

The problem is that to get the name of the columns in the dataset we need to be able to process the dataset, but at that time the dataset has not yet been uploaded to the backend.

I think to fix it we need to redesign the dataset upload.

DashAI/front/src/components/experiments/ConfigureModelsStep.jsx
maximi4321: Yes, this fix turned out to be a bit of a problem.

But in the end the one in this PR is the correct form, because the name of the DistilBert json schema changed, so that fix is no longer needed.

DashAI/front/src/components/ConfigurableObject/Inputs/ListOfStringsInput.jsx
pbadillatorrealba: I will add this in the future projects.

pbadillatorrealba (approved):

PR #107, opened 2023-06-18
Translation evaluation metrics

Oziel14:
# Summary
Classes were held for translation metrics. Classes were held for bleu and ter metrics. In addition, tests for these metrics were

## Type of change
- Back end new feature.

## How to Test

Just run test_translation_metrics.py

## Notes
Depends on the PR #106


pbadillatorrealba (approved):
Looks good

PR #108, opened 2023-06-20
Added sample method to DashAIDataset

nataliabarca:
# Summary
Added a sample method to DashAIDataset class. This method generates samples from a dataset with three different options: retrieve the first n rows, the last n rows or n random rows.

The purpose of this change is to be able to retrieve samples from a dataset when using the Prediction tab to show them to the user.

## Changes
- Added sample method to DashAIDataset in `dashai_dataset.py`
- Added new test in `test_dashai_dataset.py`

## How to Test
Run `pytest ./tests/back/dataloaders/test_dashai_dataset.py`

### DIFF
--- a/DashAI/back/dataloaders/classes/dashai_dataset.py
+++ b/DashAI/back/dataloaders/classes/dashai_dataset.py
@@ -1,7 +1,8 @@
 import json
 import os
-from typing import Dict, List
+from typing import Dict, List, Literal
 
+import numpy as np
 from datasets import ClassLabel, Dataset, DatasetDict, Value, load_from_disk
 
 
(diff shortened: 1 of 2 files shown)


DashAI/back/dataloaders/classes/dashai_dataset.py
```diff
@@ -120,6 +121,49 @@ def change_columns_type(self, column_types: Dict[str, str]):
         dataset = self.cast(new_features)
         return dataset
 
+    def sample(
+        self,
+        n: int = 1,
+        method: Literal["head", "tail", "random"] = "head",
+        seed: int | None = None,
```
pbadillatorrealba: This notation is not compatible with python 3.8 :((

You have to change it to:

```suggestion
        seed: Union[int, None] = None,
```

And import as `from typing import Union`

DashAI/back/dataloaders/classes/dashai_dataset.py
```diff
@@ -120,6 +121,49 @@ def change_columns_type(self, column_types: Dict[str, str]):
...
+    ):
```
pbadillatorrealba: Question: what kind of object returns, a python dict, dashai dataset or huggingface dataset?

Please add a return type to make this explicit

tests/back/dataloaders/test_dashai_dataset.py
```diff
@@ -94,6 +94,27 @@ def fixture_dashaidataset():
     return [datasetdict, dataloader_test]
 
 
+def test_dashaidataset_sample(dashaidataset_created: list):
```
pbadillatorrealba, line 97: Could you also test for n!=1?

pbadillatorrealba (changes requested):
Some minor comments about types and tests

nataliabarca:
All requested changes are completed 👌

pbadillatorrealba:
Nice!

I guess that merging this pr would not affect the current release, what do you think @cristian-tamblay ?

cristian-tamblay:
Good to merge!

PR #109, opened 2023-06-20
Dataset and experiment tables update

maximi4321:
# Summary

Improvements in the update of the tables (ExperimentsTable, DatasetsTable) after performing an action (delete/upload).
Better error handling in requests.
Improvements to keep the consistency between the tables.

## Type of change

- Front end new feature.
- Bug fix.
- Refactoring.

## Changes

- Created `custom/DeleteItemModal.jsx` as a generic component to confirm the deletion of an item from a table.
- Refactored the tables (`DatasetsTable.jsx` and `ExperimentsTable.jsx`) so they fetch their rows interally and save them into a state that can be updated externally using a boolean flag.
- Added buttons and changed colors to both tables to keep the consistency between them.
## How to Test

3. Follow the steps in the video in the "Screenshots" section.

## Screenshots

[table.webm](https://github.com/DashAISoftware/DashAI/assets/53657198/e6b37b46-3c2b-4215-8f9e-417a0596f5b9)


DashAI/front/src/pages/ExperimentPage.jsx
```diff
@@ -8,19 +8,22 @@ import { rows } from "../example_data/experiments";
...
         open={showNewExperimentModal}
         setOpen={setShowNewExperimentModal}
+        setUpdateTableFlag={setUpdateTableFlag}
       />
 
       {/* Experiment table */}
       <ExperimentsTable
         initialRows={rows}
         handleOpenNewExperimentModal={() => setShowNewExperimentModal(true)}
+        updateFlag={updateTableFlag}
```
pbadillatorrealba: A minor comment, but, could you use the same name for this variable in every file (`updateTableFlag`)?

DashAI/front/src/components/experiments/NewExperimentModal.jsx
```diff
@@ -90,9 +94,10 @@ export default function NewExperimentModal({ open, setOpen }) {
         newExp.name,
       );
       const experimentId = response.id;
-
+      console.log(experimentId);
```
pbadillatorrealba: Delete console log for merge versions.

DashAI/front/src/components/experiments/ExperimentsTable.jsx
```diff
@@ -82,6 +86,14 @@ function ExperimentsTable({ handleOpenNewExperimentModal }) {
     getExperiments();
   }, []);
 
+  // triggers an update of the table when updateFlag is set to true
+  React.useEffect(() => {
+    if (updateFlag) {
+      handleUpdateExperiments();
```
pbadillatorrealba: Why not to use `getExperiments` directly?

Let's leave the notation of handlers only to handle changes caused by direct user action.

pbadillatorrealba: And another question, there is some possibility that `getExperiments` fail and updateFlag always be maintained as true?
I think it would be better first to change the flag and then to do the call.

DashAI/front/src/components/custom/DeleteItemModal.jsx
```diff
@@ -0,0 +1,52 @@
...
+    <React.Fragment>
+      <GridActionsCellItem
+        key="delete-button"
+        icon={<DeleteIcon />}
+        label="Delete"
+        onClick={() => setOpen(true)}
+        sx={{ color: "#ff8383" }}
+      />
+      <Dialog open={open} onClose={() => setOpen(false)}>
+        <div>
```
pbadillatorrealba: Is this `div` necessary?

DashAI/front/src/components/custom/DeleteItemModal.jsx
```diff
@@ -0,0 +1,52 @@
...
+    setOpen(false);
+  };
+  return (
+    <React.Fragment>
+      <GridActionsCellItem
+        key="delete-button"
+        icon={<DeleteIcon />}
+        label="Delete"
+        onClick={() => setOpen(true)}
+        sx={{ color: "#ff8383" }}
```
pbadillatorrealba: I think it would be better to use the colors that are included in the ui material, so if we change one of these it will not be necessary to look for them in the code.

https://mui.com/system/palette/#color

```suggestion
        sx={{ color: "error.main" }}
```

pbadillatorrealba (changes requested):
Nice refactor!

I left some minor comments.

DashAI/front/src/components/custom/DeleteItemModal.jsx
maximi4321: Thank you, i also used mui colors with another component that had a similar problem.
A summary of the changes:
- I updated `DeleteItemModal.jsx` to use "error.main" like in the suggested change.
- Similarly  i updated `EditDatasetModal.jsx` to use "warning.main".
- I added values to `style/theme.js` to define custom colors for "error.main" and "warning.main"

pbadillatorrealba (approved):

PR #114, opened 2023-07-03
Experiment Modal improvement

maximi4321:
# Summary

The names of the form fields were modified in the "Configure models" step, and the background color was removed from the parameter form.
## Type of change

- Bug fix.
- Refactoring.

## Changes

- The form field names have been updated in the "Configure models" step to align with the names they receive in the runs table. "Nickname" has been changed to "Name" and "Type" has been changed to "Model"
-  The Paper component that previously contained the parameter form has been replaced with Box, so the background color now is the same of the modal.

## How to Test

3. Go to Experiments tab and create a new experiment.
4. In step 3 "Configure models" try adding a model and configuring it.

## Screenshots

1. **Name changes:** The previous "Nickname" field has been renamed to "Name" and "Type" has been renamed to "Model". These changes are applied to both the labels of the form fields and the table.

Name change

2. **Parameter form background color is the same of the modal**

ParamForm (1)


pbadillatorrealba (approved):
That's great, I've been thinking about changing nickname for a while 😂.
Approved.

PR #115, opened 2023-07-05
Job Queue

Riul1999:
# Summary

Added new Jobs architecture to handle asynchronous processes (runner).

## Type of change

- Back end new feature.
- Refactoring.

## Changes

- Added `job` endpoint to handle Jobs:
  - GET `/job/` to get all the pending jobs in the jobs queue.
  - GET `/job/{job_id}` to get the job with job_id from the jobs queue.
  - POST `/job/runner/` to create a runner job.
  - DELETE `/job/{job_id}` to cancel the job wuth job_id from the jobs queue.
- Delete `runner` endpoint, now it is on the `job` endpoint.
- Update exec_runner function, now it only receives the run_id and db. Also add documentation an error handling.
- Refactor test_runner into test_jobs to take into account the changes made.
- Added Job pydantic model, to abstract a job to execute.
- Added BaseJobQueue ADT, to abstract the usage of the job queue:
  - `put` adds a Job into the queue and returns its id.
  - `get` extracts a job from the queue.
  - `async_get` extracts a job from the queue, if the queue is empty waits until a job arrieves.
  - `peek` retrieves a job from the queue without extracting it.
  -  `is_empty` is a predicate that indicates if the queue is empty.
  - `to_list` transforms the queue in a list of Jobs.
- Added SimpleJobQueue implementation of BaseJobQueue, it uses the async Queue provided by asyncio.
- Added job_queue into `core/config.py`.
- Added `job_queue_loop` function to retrieves jobs from the queue and execute it.
- Added `test_simple_job_queue.py` to test the SimpleJobQueue implementation.
- Added `pytest-asyncio` to the `requirements_dev.txt` file, it's used to test the `async_get` method of the SimpleJobQueue class.

## How to Test

- To test the SimpleJobQueue class run `pytest .\tests\back\job_queue\test_simple_job_queue.py`.
- To test the job endpoint run `pytest .\tests\back\api\test_jobs.py`


DashAI/back/core/runner.py
```diff
@@ -19,50 +20,129 @@ class RunnerError(Exception):
...
 
-def execute_run(
-    dataset: Dict[str, DashAIDataset],
-    task: BaseTask,
-    model: BaseModel,
-    metrics: List[BaseMetric],
-    run: Run,
-    db: Session,
-):
+def execute_run(run_id: int, db: Session):
```
pbadillatorrealba, line 24: This function is too complex. 
I think that in the future we should refactor it and separate it by steps, where each step has its own private method.

DashAI/back/core/runner.py
```diff
@@ -19,50 +20,129 @@ class RunnerError(Exception):
...
+    """Function to train and evaluate a Run.
+    It retrieves all the objects associated with the run and then it:
+    - Trains the model.
+    - Evaluate the model.
+    - Save the trained model.
```
pbadillatorrealba, line 29: If it is refactored into methods, perhaps the code is self-explanatory and this part of the documentation would not need to be updated.

DashAI/back/job_queues/base_job_queue.py
```diff
@@ -0,0 +1,136 @@
...
+
+    def to_dict(self) -> dict:
+        """Returns a dict representation of the Job.
+
+        Returns
+        ----------
+        dict
+            dictionary with most of the fields of the Job.
+        """
+        return {"id": self.id, "type": self.type, "kwargs": self.kwargs}
```
pbadillatorrealba: Externalize the pydantic model to its own file, perhaps in another folder such as `core/schemas/job_model.py`.

pbadillatorrealba: Include the enum in this refactorization.

DashAI/back/job_queues/base_job_queue.py
```diff
@@ -0,0 +1,136 @@
...
+
+
+class JobQueueError(Exception):
+    """Exception raised when a method of the job queue fails."""
```
pbadillatorrealba: Externalize this exception to its own module, perhaps in another folder such as `core/exceptions/job_exceptions.py`.

DashAI/back/job_queues/simple_job_queue.py
pbadillatorrealba, line 1: Perhaps to make the basis of the implementation more explicit  it would be better to call it `BuiltInAsyncioJobQueue` instead of `SimpleJobQueue`?

pbadillatorrealba: If so, remember to rename the file.

DashAI/back/job_queues/simple_job_queue.py
```diff
@@ -0,0 +1,96 @@
...
+
+from DashAI.back.job_queues.base_job_queue import BaseJobQueue, Job, JobQueueError
+
+
+class SimpleJobQueue(BaseJobQueue):
+    """JobQueue implementation using asyncio Queue."""
+
+    queue: Queue = Queue()
+
+    def _search_and_split(self, job_id: int) -> tuple[List[Job], Job, List[Job]]:
```
pbadillatorrealba: Use Tuple to ensure python 3.8 compatibility.

```suggestion
    def _search_and_split(self, job_id: int) -> Tuple[List[Job], Job, List[Job]]:
```

DashAI/back/job_queues/simple_job_queue.py
```diff
@@ -0,0 +1,96 @@
...
+        """Split the queue using the job with id job_id as a pivot.
+
+        Parameters
+        ----------
+        job_id: int
+            Id of the Job used to split the queue.
+
+        Returns
+        ----------
+        tuple(List Job, Job, List Job)
```
pbadillatorrealba, line 23:
```suggestion
         Tuple[List[Job], Job, List[Job]]
```

DashAI/back/job_queues/simple_job_queue.py
```diff
@@ -0,0 +1,96 @@
...
+        return (first_part, target_job, second_part)
+
+    def put(self, job: Job) -> int:
+        job.id = uuid.uuid4().int
+        self.queue.put_nowait(job)
+        return job.id
+
+    def get(self, job_id: int | None = None) -> Job:
+        if self.is_empty():
+            raise JobQueueError(f"Searching in Empty Queue for job {job_id}")
```
pbadillatorrealba:
```suggestion
            raise JobQueueError(f"Error trying to get job {job_id}: the async queue is empty.")
```

DashAI/back/api/api_v1/endpoints/jobs.py
```diff
@@ -0,0 +1,172 @@
...
+from DashAI.back.job_queues import Job, JobQueueError, JobType
+
+logging.basicConfig(level=logging.DEBUG)
+log = logging.getLogger(__name__)
+
+router = APIRouter()
+
+
+@router.post("/start/")
+async def start_job_queue(
```
pbadillatorrealba, line 23: I seriously think we should change the background task to some other type of mechanism and launch it once the application starts, not when this method is invoked. 

Maybe this tutorial should work: https://johnsturgeon.me/2022/12/10/asyncio-queue

DashAI/back/api/api_v1/endpoints/jobs.py
```diff
@@ -0,0 +1,172 @@
...
+    except JobQueueError as e:
+        log.exception(e)
+        raise HTTPException(
+            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+            detail="Internal job queue error",
+        ) from e
+    return job
+
+
+@router.post("/runner/", status_code=status.HTTP_201_CREATED)
```
pbadillatorrealba, line 87: I imagine that there must be some good reason, but, why here is not used directly put to `/`?

DashAI/back/core/job_queue.py
```diff
@@ -0,0 +1,30 @@
...
+        boolean to set the while loop condition.
+    """
+    while not job_queue.is_empty() if stop_when_queue_empties else True:
+        try:
+            job: Job = await job_queue.async_get()
+            job.func(**job.kwargs)
+        except exc.SQLAlchemyError as e:
+            log.exception(e)
+        except RunnerError as e:
+            log.exception(e)
```
pbadillatorrealba, line 30: I have my doubts about the background tasks end up blocking or not the main thread. 

According to what I read, background tasks adds the task to the task loop of the same thread. 
So, if the training function takes too long, it would still block the execution since it would not release the thread at any time.
This would mean that the server would stop working.

I think that at this point it would be necessary to make some mechanism that:

1. Get the process.
2. Launch a different process that executes the training.
3. Wait asynchronously for the process to finish.
4. Continue with the next task.
5. If there is no task left, wait asynchronously until the next one.


```suggestion

import asyncio
import logging
from multiprocessing import Process

from DashAI.back.core.config import job_queue
from DashAI.back.job_queues import Job

logging.basicConfig(level=logging.DEBUG)
log = logging.getLogger(__name__)


async def job_queue_loop(stop_when_queue_empties: bool):
    """Loop function to execute all the pending jobs in the job queue.
    If the the param stop_when_queue_empties is True, the loop returns when
    the queue empties, else it waits until  new jobs come in.
    Parameters
    ----------
    stop_when_queue_empties: bool
        boolean to set the while loop condition.
    """
    try:
        while True:
            
            # if there is no enqueued job, wait.
            try: 
                job: Job = job_queue.get()
            except JobQueueError:
                asyncio.sleep(1)
            
            # else, execute the train.
            train_process = Process(target=job.func, args=job.kwargs, daemon=True)
            pid = train_process.start()

            while train_process.is_alive():
                log.debug(
                    f"Awaiting {job.id} process (pid={pid}) process for 1 second."
                )
                asyncio.sleep(1)  # sleep for 1 seconds.

            result = train_process.join()
            # do something with the result...

            

        # the child process should care about this exceptions...
        # except exc.SQLAlchemyError as e:
        #     log.exception(e)
        # except RunnerError as e:
        #     log.exception(e)

    except Exception as e:
        msg = f"Error in the training execution loop: {e}.\nShutting down the app."
        log.exception(msg)
        raise RuntimeError(msg) from e

```

Even going a little further, maybe this loop should be called something like `MultiprocessingJobQueueLoop`.
Finally, you have to try this. I'm not sure if it works or not.

pbadillatorrealba: If this goes well, I think this should be unified with the asynchronous queue start. Perhaps this could be an asynchronous method of the same `SimpleJobQueue` class.

DashAI/back/job_queues/simple_job_queue.py
pbadillatorrealba: Also, if the change to multiprocessing goes well too, then perhaps the name `MultiprocessingAsyncJobQueue` would be better.

pbadillatorrealba (changes requested):
The changes are really great!

I think we are going in the right direction in making the queues.
However, I have doubts that everything will work well with the background tasks. 

See the comments I left in the job_queue file. I think that summarizes all the doubts plus an implementation that could quickly get us out of trouble.

I also left some comments about the API, code, typing and documentation.

pd: this is probably the hardest pr I've ever had to review in my life 😂.

DashAI/back/api/api_v1/endpoints/jobs.py
Riul1999: This is because there will be (in future versions) more than one type of work, concretely the prediction work.

pbadillatorrealba (approved):

PR #116, opened 2023-07-07
Minor frontend fixes

maximi4321:
# Summary

This PR includes small changes that were pending to be implemented in the frontend.

## Type of change

- Bug fix.
- Refactoring.

## Changes

1. Fixed the responsiveness in `ResultsPage.jsx`. The drawer no longer covers the table even if the screen width is changed.
2. The browser's autofill functionality has been disabled for text input fields, because React has some [problems handling autofill on controlled components](https://github.com/facebook/react/issues/1159)
3. A bug in the table for selecting runs to execute has been fixed (`RunnerDialog.jsx`). Now the user's selection is preserved (previously, all runs were reselected on any update).
4. The modals for confirming the deletion of an item in the tables have been refactored. Since they were all very similar to each other, they have been replaced with a generic component.
5. The API for editing datasets has been restored, which previously was not working due to compatibility issues with the backend.

## How to Test

3. Follow the steps in the "Screenshots" section.

## Screenshots

<details>
<summary>Click to toggle the screenshots</summary>
1. Responsiveness in results page.

responsiveness

2. Browser autofill is disabled

autofill

3. Keeps the users run selection

[Select.webm](https://github.com/DashAISoftware/DashAI/assets/53657198/d08c8b6f-20b7-410f-bfc5-87c0aae2ad3d)

4. Generic modal to delete item

generic modal

5. Restored edit dataset API

[edit dataset.webm](https://github.com/DashAISoftware/DashAI/assets/53657198/bf5ae3e9-9829-49f3-b863-cc92ac4f7303)
</details>


DashAI/front/src/api/datasets.ts
pbadillatorrealba, line 1: Why not to use the axios interface to generate the query string?

DashAI/front/src/components/results/RunsTable.jsx
```diff
@@ -192,7 +194,16 @@ function RunsTable({ experimentId }) {
...
+    <Paper
+      sx={{
+        p: 4,
+        [theme.breakpoints.up("xs")]: { ml: 8 },
+        [theme.breakpoints.up("sm")]: { ml: 12 },
+        [theme.breakpoints.up("md")]: { ml: 20 },
+        [theme.breakpoints.up("lg")]: { width: "80vw", ml: 15 },
+        [theme.breakpoints.up("xl")]: { width: "80vw", ml: 10 },
+      }}
+    >
```
pbadillatorrealba: Question: Is the margin left assigned from the left edge of the image? I think it would be better if it was with respect to the drawer.

DashAI/front/src/components/experiments/RunnerDialog.jsx
```diff
@@ -92,8 +83,19 @@ function RunnerDialog({ experiment, expRunning, setExpRunning }) {
...
-        console.error("Unkown Error", error.message);
+        console.error("Unknown Error", error.message);
       }
+    } finally {
+      enqueueSnackbar(`${experiment.name} has finished running`, {
+        variant: "info",
+        anchorOrigin: {
+          vertical: "top",
+          horizontal: "right",
+        },
```
pbadillatorrealba, line 87: According to snackbars pr, this is no longer necessary.

DashAI/front/src/components/experiments/RunnerDialog.jsx
```diff
@@ -32,15 +32,17 @@ function RunnerDialog({ experiment, expRunning, setExpRunning }) {
     setLoading(true);
     try {
       const runs = await getRunsRequest(experiment.id.toString());
-      const runInExecution = runs.find((run) => run.status === 2); // search a run with status "running"
+      const runInExecution = runs.find((run) => run.status === 2); // searches for a run with the status "running"
```
pbadillatorrealba:
```suggestion
      const runsInExecution = runs.find((run) => run.status === 2); // searches for a run with the status "running"
```

DashAI/front/src/components/experiments/RunnerDialog.jsx
```diff
@@ -92,8 +83,19 @@ function RunnerDialog({ experiment, expRunning, setExpRunning }) {
...
       }
+    } finally {
+      enqueueSnackbar(`${experiment.name} has finished running`, {
+        variant: "info",
+        anchorOrigin: {
+          vertical: "top",
+          horizontal: "right",
+        },
+      });
+      setExpRunning({ ...expRunning, [experiment.id]: false });
```
pbadillatorrealba, line 89: What does `[experiment.id]: false ` do?

pbadillatorrealba (commented):
The changes in general are fine!
I left some comments and questions.

DashAI/front/src/components/experiments/RunnerDialog.jsx
maximi4321: This was fixed by merging with staging.

DashAI/front/src/components/experiments/RunnerDialog.jsx
maximi4321: That line searches if there is any run currently running.

I used "run" instead of "runs" because the `find` method returns the first element in the array that satisfies the condition.

DashAI/front/src/components/experiments/RunnerDialog.jsx
maximi4321: `expRunning` is a state that contains information about the execution status of all experiments in the table.

What that line does is indicate that the experiment has finished running, whether it ended successfully or ocurred an error.

DashAI/front/src/components/results/RunsTable.jsx
maximi4321: Yes, the margin left is assing from the edge.

I am trying to define the margin with respect to the drawer, and I think I managed to do it. But by modifying that, now i haven't been able to make the table take up the entire space.

DashAI/front/src/components/experiments/RunnerDialog.jsx
pbadillatorrealba: Then, maybe call this variable `firstRunInExecution`?

DashAI/front/src/components/results/RunsTable.jsx
pbadillatorrealba: I think that Drawer is not the best component to use in this case (it is more intended to be the drawer where the navigation menu of the page is located) but List.

This will allow you to have the objects live in the same view and therefore, the size calculations are done with respect to the list and the table.

Suggestion: You could implement a view with two `Grid`, the one on the right with a small size where the `List` of experiments (`ListItems`) goes and the one on the left where the visualization goes. 
And then make the grid sizes variable according to the size of the page using the typical sm, md, lg, etc...

DashAI/front/src/pages/ResultsPage.jsx
```diff
@@ -1,19 +1,60 @@
...
+      <Grid
+        item
+        sx={{
+          backgroundColor: "#212121",
+          height: `calc(100vh - ${appBarHeight}px)`,
+        }}
+      >
+        <ExperimentsList />
+      </Grid>
+      <Grid item sx={{ my: 5 }}>
```
pbadillatorrealba: Add a little margin on right.

```suggestion
      <Grid item sx={{ my: 5, mr: 2 }}>
```

pbadillatorrealba (changes requested):
It looks good!

I think Container (in `app.jsx`) is the component that is causing the strange behaviors regarding the view responsive sizes.
For now, it's fine, but in the future we will have to think about how to deactivate it.

A minor margin change was requested.

Question: Could you disable the black background of the experiment list?

pbadillatorrealba (approved):

PR #121, opened 2023-07-14
Back/predict endpoint

nataliabarca:
# Summary

Implemented predict endpoint to perform model prediction for a particular run, given some sample values. 
This endpoint works for classification tasks.

## Type of change

* Backend new feature.

## Changes

* New endpoint in the `predict.py` file
  *  The POST endpoint is used to perform a prediction using a specific run and given sample data, passed as a file.
  * GET, DELETE and PATCH endpoint only return a 501NotImplemented HTTP Error.
* Test `test_predict.py`
  * The `test_make_prediction` checks if the endpoint returns a list with the same number of elements as the given input samples.

## How to Test

Run `pytest tests/back/api/test_predict.py`


tests/back/api/test_predict.py
```diff
@@ -0,0 +1,126 @@
...
+            },
+            {
+                "SepalLengthCm": 100,
+                "SepalWidthCm": 100,
+                "PetalLengthCm": 100,
+                "PetalWidthCm": 100,
+            },
+        ]
+    }
+    response = client.post(f"""/api/v1/predict/?run_id={run_id}""", json=data)
```
cristian-tamblay: This is a problem we have carried for a while, but POST request with query parameters in URL have no sense, all parameters should be included in the body of the request. Dataset POST endpoint is the only one doing this correctly, and the way to do it is using pydantic. I'm going to fix the other endpoints' POSTs, but i prefer to avoid adding more technical debt to the software

cristian-tamblay (approved):

PR #128, opened 2023-07-17
Fix Metrics in Runner

Riul1999:
# Summary

Change the way of obtaining the metrics used in the runner.

## Type of change

- Bug fix.

## Changes

- Obtain the metrics using the structure used in the components endpoint.
- Create test to check the calculation of the metrics.

## How to Test

To test the change run pytest `.\tests\back\api\test_runner.py`

### DIFF
--- a/tests/back/api/test_runner.py
+++ b/tests/back/api/test_runner.py
@@ -199,8 +199,10 @@ def test_exec_runs(client: TestClient, run_id: int):
     response = client.get(f"/api/v1/run/{run_id}")
     data = response.json()
     assert isinstance(data["train_metrics"], dict)
-    assert isinstance(data["validation_metrics"], dict)
-    assert isinstance(data["test_metrics"], dict)
+    assert "DummyMetric" in data["train_metrics"]
+    assert data["train_metrics"]["DummyMetric"] == 1
+    assert data["train_metrics"] == data["validation_metrics"]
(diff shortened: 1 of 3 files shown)


tests/back/api/test_runner.py
```diff
@@ -199,8 +199,10 @@ def test_exec_runs(client: TestClient, run_id: int):
     response = client.get(f"/api/v1/run/{run_id}")
     data = response.json()
     assert isinstance(data["train_metrics"], dict)
-    assert isinstance(data["validation_metrics"], dict)
-    assert isinstance(data["test_metrics"], dict)
+    assert "DummyMetric" in data["train_metrics"]
+    assert data["train_metrics"]["DummyMetric"] == 1
+    assert data["train_metrics"] == data["validation_metrics"]
+    assert data["train_metrics"] == data["test_metrics"]
```
cristian-tamblay, line 205: These shouldn't be always equal, we are calculating metrics in different splits, which should have different results

cristian-tamblay (changes requested):
We also should add these changes from my previous PR on this subject
image

tests/back/api/test_runner.py
Riul1999: But the DummyMetric used in the test always returns 1, so the result is independent of the data.

cristian-tamblay (approved):

PR #131, opened 2023-07-18
Text classification working from the front end

Oziel14:
# Summary

Changes were made to the registry to make the text classification model operative from the front end, in addition to changing the type of average used in the metrics to vary the values. Also a change was made to the dataloaders since an obsolete class was being used.

## Type of change

- Bug fix.
- Refactoring.

## How to Test

The old tests still work and a small change was made to the text classification models test to deliver an outline as the user would send it.

## Notes

This PR must be approved before that of PR #103

### DIFF
--- a/DashAI/front/src/components/experiments/ConfigureModelsStep.jsx
+++ b/DashAI/front/src/components/experiments/ConfigureModelsStep.jsx
@@ -45,7 +45,12 @@ function ConfigureModelsStep({ newExp, setNewExp, setNextEnabled }) {
 
   const getModelSchema = async () => {
     try {
-      const schema = await getModelSchemaRequest(selectedModel);
+      let schema;
+      if (selectedModel === "DistilBertTransformer") {
+        schema = await getModelSchemaRequest("DistilBERT");
+      } else {
+        schema = await getModelSchemaRequest(selectedModel);
+      }
       return schema;
     } catch (error) {
       enqueueSnackbar("Error while trying to obtain model schema");
--- a/DashAI/front/src/components/experiments/EditModelDialog.jsx
+++ b/DashAI/front/src/components/experiments/EditModelDialog.jsx
@@ -33,7 +33,13 @@ function EditModelDialog({
   const getObjectSchema = async () => {
     setLoading(true);
     try {
-      const schema = await getModelSchemaRequest(modelToConfigure);
+      // const schema = await getModelSchemaRequest(modelToConfigure);
+      let schema;
+      if (modelToConfigure === "DistilBertTransformer") {
+        schema = await getModelSchemaRequest("DistilBERT");
+      } else {
+        schema = await getModelSchemaRequest(modelToConfigure);
+      }
       setModelSchema(schema);
     } catch (error) {
       enqueueSnackbar("Error while trying to obtain model schema");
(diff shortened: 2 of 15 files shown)


DashAI/front/src/components/experiments/ConfigureModelsStep.jsx
```diff
@@ -45,7 +45,12 @@ function ConfigureModelsStep({ newExp, setNewExp, setNextEnabled }) {
 
   const getModelSchema = async () => {
     try {
-      const schema = await getModelSchemaRequest(selectedModel);
+      let schema;
+      if (selectedModel === "DistilBertTransformer") {
+        schema = await getModelSchemaRequest("DistilBERT");
+      } else {
+        schema = await getModelSchemaRequest(selectedModel);
+      }
```
pbadillatorrealba, line 53: This looks like a future bug. 
It is not possible to no hardcode `DistilBERT`?
If not, add a card indicating that we should solve this problem.

DashAI/front/src/components/experiments/EditModelDialog.jsx
```diff
@@ -33,7 +33,13 @@ function EditModelDialog({
...
     setLoading(true);
     try {
-      const schema = await getModelSchemaRequest(modelToConfigure);
+      // const schema = await getModelSchemaRequest(modelToConfigure);
+      let schema;
+      if (modelToConfigure === "DistilBertTransformer") {
+        schema = await getModelSchemaRequest("DistilBERT");
+      } else {
+        schema = await getModelSchemaRequest(modelToConfigure);
+      }
```
pbadillatorrealba, line 42: Same as before, it is possible not to hardcode this?

tests/back/models/test_text_class_models.py
```diff
@@ -44,17 +44,9 @@ def fixture_load_dashaidataset():
...
-    distilbert.fit(load_dashaidataset["train"])
-    return distilbert
-
-
-@pytest.fixture(scope="session", name="model_fit_with_params")
 def text_class_model_fit_with_params(load_dashaidataset: DatasetDict):
-    distilbert = DistilBertTransformer(
-        num_train_epochs=2, per_device_train_batch_size=32
-    )
+    params = {"num_train_epochs": 2, "batch_size": 32, "device": "cpu"}
```
pbadillatorrealba: Can be one training epoch (to avoid saturating the CI)?

tests/back/models/test_text_class_models.py
```diff
@@ -77,7 +69,8 @@ def test_predict_text_class_model(
 
 
 def test_not_fitted_text_class_model(load_dashaidataset: DatasetDict):
-    distilbert = DistilBertTransformer()
+    params = {"num_train_epochs": 2, "batch_size": 32, "device": "cpu"}
```
pbadillatorrealba: Same here, could `num_train_epochs` be changed to 1?

pbadillatorrealba (commented):
I think the changes are fine :)
But I have doubts about a couple of points, which I left pointed out.

DashAI/front/src/components/experiments/ConfigureModelsStep.jsx
pbadillatorrealba: It will be fixed in another pr.

pbadillatorrealba (approved):

PR #136, opened 2023-07-27
Fixed multiple schemas with problems

cristian-tamblay:
# Summary

Fixed some model schemas that had problems

## Type of change

- Bug fix.

## Changes

- Fixed HistGradientBoosting l2_regularization from string to float
- Fixed some wrong descriptions on dummyclassifier
- Fixed criterion on DecisionTree

## How to Test

Now its possible to train sucessfully a DecisionTreeClassifier and a HistGradientBoosting

## Screenshots

image

### DIFF
--- a/DashAI/back/models/parameters/models_schemas/DecisionTreeClassifier.json
+++ b/DashAI/back/models/parameters/models_schemas/DecisionTreeClassifier.json
@@ -6,11 +6,11 @@
     "criterion": {
       "oneOf": [
         {
-          "error_msg": "The 'penalty' parameter should be one of 'gini' or 'entropy'.",
-          "description": "The function to measure the quality of a split. Supported criteria are “gini” for the Gini impurity and “entropy” for the Shannon information gain.",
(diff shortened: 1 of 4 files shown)


DashAI/back/models/parameters/models_schemas/DecisionTreeClassifier.json
```diff
@@ -53,7 +53,7 @@
           "error_msg": "The 'max_features' parameter should be one of 'auto', 'sqrt', 'log2', 'null'.",
           "description": "The number of features to consider when looking for the best split",
           "type": "string",
-          "default": "l2",
+          "default": null,
           "enum": ["auto", "sqrt", "log2", null]
```
maximi4321, line 57: Adding "null" in "enum" causes an empty option to exist in the dropdown of the parameter form.

If that parameter is configured, the default value cannot be re-selected.

maximi4321:
The changes work well for me, I was able to run the models both with their default values and by configuring their parameters.

I left a small comment with a parameter that it would be better not to modify in case of opening the parameter form.

maximi4321 (approved):

PR #139, opened 2023-08-03
Pydantic post endpoints

cristian-tamblay:
# Summary

Fixing the usage of query parameters on POST requests. Also doing some minor refactoring in pydantic schemas and endpoints 

## Type of change

- Bug fix.
- Refactoring.

## Changes

- Enabled pydantic into api's POST requests.
- Refactored dataloader pydantic models into `api_v1/endpoints/schemas`
- Moved general pydantic validation into `api_v1/utils.py`

## How to Test

POST requests should contain its data inside the form instead of query parameters

## Note
Actually tests are not passing because they were using query params instead of params in body. Fixing in next pushes


DashAI/back/api/api_v1/schemas/datasets_params.py
pbadillatorrealba, line 1: Is it a good idea for these things to have default values?
I'm leaning towards them not having them or not at least explicitly in the models.
Having them there may give rise to default values that later we will not know where they are coming from.

pbadillatorrealba (commented):
Great, I think this change will standardize the theme of the post finally!

I have a question that could potentially involve all the changes: why use Form? Couldn't you just use the Body of the request in json format?
Then we could directly use the pydantic models without using the util function parsing.

I also left a comment on the pydantic models.

pbadillatorrealba:
According to the documentation and discussions I have read, in the Front, there is no difference in performance and data transmission between formdata and json. 

The "small" difference is that formdata is oriented to pages without js, while json, being manipulable by js, is the preferred option in this type of environment (react for example).

Regardless of what we choose, I think the important thing is consistency.

- https://stackoverflow.com/questions/8604717/json-vs-form-post
- https://www.reddit.com/r/learnprogramming/comments/v4jgec/which_is_efficient_formdata_or_json_to_send_data/
- https://www.vinaysahni.com/best-practices-for-a-pragmatic-restful-api#json-responses

PR #142, opened 2023-08-10
Accept null values in parameter form

maximi4321:
# Summary

This pr includes modifications to the json files of configurable objects and to the form rendering so that null values are accepted when indicated.

## Type of change

- Bug fix.

## Changes

- The json of the configurable objects are modified to indicate when a parameter accepts null values.
The type must be either a string or an array containing the string "null" and the type.
```
{
  ...
  "type": ["integer", "null"]
  ...
}
```
- The rendering of the configurable object form and its validation are adapted to the changes in the json files.
- In the json files of configurable objects, the type "float" is replaced by "number" and this change is taken into account in the rendering and validation.
- Some warnings have been fixed
  - A boolean is now correctly used instead of a string in the `error` property of the TextFields
  - TextFields should not use null values, so the empty string is used to work with the TextField and is converted to null when sent to the backend.

## How to Test

3. Create an experiment, configure its parameters and try to run it.


cristian-tamblay (approved):
Tested and working

PR #143, opened 2023-08-11
Small fix on the results page

maximi4321:
# Summary

By deactivating the container on the results page, the specific view of a run was affected. This PR fixes this problem

## Type of change

- Bug fix.

## Changes

- A container (`CustomLayout`) is added in the specific view of runs to restore the layout as it was before.

## How to Test

## Screenshots

results


cristian-tamblay (approved):
Works as intended

PR #148, opened 2023-08-21
Front/fixes

maximi4321:
# Summary

This PR improves the alignment of the components of the parameter form and adds more information to the button to start an experiment.

## Type of change

- Bug fix.
- Refactoring.

## Changes

- A generic component (`FormInputWrapper.jsx`) is created that wraps each input in a `Grid` that arranges the layout by vertically centering the position of each iconButton to display tooltips with description.
- The text and icon of the button to start an experiment is changed to "finished" once all the runs have finished running.
- Change the capitalization in the "ConfigurableObject" folder to "configurableObject" as in the rest of the folders.
- In tables where items are added, instead of being added at the end, they are added at the start.
- Tables update is fixed, now they are updated even if the request takes too much time
- fixed warnings
## How to Test

3. Test the views shown in the screenshots section

## Screenshots

1. Parameter Form

paramForm

2. Start experiment button 

finished

3. Add items at the start instead of at the end

[add.webm](https://github.com/DashAISoftware/DashAI/assets/53657198/df3be2e9-0d8b-4df3-bcee-ed1b0f554c72)


DashAI/front/src/components/configurableObject/Inputs/FormInputWrapper.jsx
```diff
@@ -0,0 +1,39 @@
...
+ * @param {React.ReactNode} children the input component to render in the layout
+ *
+ */
+function FormInputWrapper({ name, description, children }) {
+  return (
+    <Grid
+      container
+      direction="row"
+      key={name}
+      sx={{ display: "flex", alignItems: "center" }}
```
pbadillatorrealba: Why not to use the pre-defined parameters like `alignItems="center"`

```suggestion
      alignItems="center"
```

DashAI/front/src/components/configurableObject/Inputs/FormInputWrapper.jsx
```diff
@@ -0,0 +1,39 @@
...
+      spacing={1}
+    >
+      {/* Form input */}
+      <Grid item xs={11}>
+        {children}
+      </Grid>
+
+      {/* Tooltip with the parameter description */}
+      <Grid item xs={1} sx={{ mb: 2 }}>
```
pbadillatorrealba, line 21: This margin bottom affects the complete row or just the tooltip?
If it is the first case, maybe it could be better in the container specify a margin between components (check [`rowSpacing` and `columnSpacing`](https://mui.com/material-ui/react-grid/#row-amp-column-spacing) properties)

DashAI/front/src/components/datasets/DatasetModal.jsx
```diff
@@ -64,6 +64,7 @@ function DatasetModal({ open, setOpen, updateDatasets }) {
       formData.append("file", newDataset.file);
       await uploadDatasetRequest(formData);
       enqueueSnackbar("Dataset uploaded successfully", { variant: "success" });
+      updateDatasets();
```
pbadillatorrealba, line 67: This function is to fetch the datasets?
Maybe if it does that, then it would be good to put getDatasets instead of updateDatasets.


If so, what happens if this call fails?
Maybe there should be an exclusive snackbar for this?

It could be that you can pull out which call failed from the errors to handle them by error type.

DashAI/front/src/components/experiments/ExperimentsTable.jsx
```diff
@@ -191,6 +197,8 @@ function ExperimentsTable({
             },
           },
         }}
+        sortModel={[{ field: "id", sort: "desc" }]}
+        columnVisibilityModel={{ id: false }}
```
pbadillatorrealba, line 201: Is this to have the ids of each row when handling an action?

DashAI/front/src/components/experiments/NewExperimentModal.jsx
```diff
@@ -93,10 +98,10 @@ export default function NewExperimentModal({
       const experimentId = response.id;
       await uploadRuns(experimentId);
 
-      setUpdateTableFlag(true);
       enqueueSnackbar("Experiment successfully created.", {
         variant: "success",
       });
+      updateExperiments();
```
pbadillatorrealba, line 104: Same comment of the fetchs and errores here.

DashAI/front/src/components/datasets/DatasetModal.jsx
pbadillatorrealba: I'm seeing that the states are getting pretty nested (I saw about 3 components before I knew where the datasets were).

I think it's a good time to consider having centralized states with redux or something like that.
Would you be interested in discussing it?

pbadillatorrealba (commented):
I left some comments regarding MUI and some that I think are interesting regarding app data and state handling.

DashAI/front/src/components/datasets/DatasetModal.jsx
maximi4321: The function looks like this: `updateDatasets={() => setUpdateTableFlag(true)}`

What it does is change a state to notify the component that contains the table to make a request with getDatasets.
In the component that contains the table is where error handling is done and a snackbar is used to notify the user.

DashAI/front/src/components/experiments/ExperimentsTable.jsx
maximi4321: I added the id to sort the rows so that when a new experiment is added it will always be positioned at the beginning of the table.

DashAI/front/src/components/configurableObject/Inputs/FormInputWrapper.jsx
maximi4321: It affects the grid, but that grid contains only the tooltip.

I tried to use `columnSpacing` and `rowSpacing` but i couldn't make it work, I think it is because there is only one tooltip inside each grid.

pbadillatorrealba (approved):

PR #149, opened 2023-08-23
Dataloaders and dataset general refactor

pbadillatorrealba:
# Summary

This pr is focused on doing a general refactoring of the dataloaders and dataset module, including improvements in method interfaces, typing, documentation, tests and others.
 In addition, beartype is included in most beartype functions to force type checking and avoid possible future bugs.

## Type of change

- Refactoring.
- Documentation.
- Bug fix.

## Changes

- Improved load_data functions. Now they no longer do type checking (beartype does it underneath).
- On the other hand, the file and url parameters were replaced by filepath_or_buffer, which accepts both types and infers the behavior according to the type of object delivered (filepath_or_buffer comes from how pandas defines this type of parameters, I think it is good to follow that convention).
- Many dataloader tests (and tests of other modules that depend on them) were modified to improve readability, however, their functionality should not have been altered. In other words, they should still be testing the same.
- Added types to most dashai_dataset methods. In addition, these additions include verification through beartype.
- Several error messages were improved.
- Many documentations were improved and formatted in numpy style.

## How to Test

Install the latest packages with pip

```bash
$ pip install -U requirements.txt
```

Run all the tests as usual
```bash
$ pytest .
```

## Notes

- This pr depends on #146 
- I recommend first checking the dataloaders, then dataset and dataloader tests and the remaining tests.


pbadillatorrealba:
@cristian-tamblay , updated to the latest staging version.

cristian-tamblay (approved):
These changes are great, i wish there was a way to have shorter PRs when doing this

PR #153, opened 2023-09-14
 BaseConverter and DropColumns

Lucas-CE:
# Summary

1. A new class representing the base class for preprocessors has been created. This class provides an abstract process function.

2. Furthermore, a new class has been introduced to represent the Column Removal Preprocessor. This class includes a process method that removes columns within the specified initial and final index range provided as input to the function.

## Type of change

- Back and new feature

## How to Test

- The tests are not ready


DashAI/back/preprocessor/base_preprocessor.py
```diff
@@ -0,0 +1,24 @@
+from abc import abstractmethod
+from typing import Final
+
+
+class BasePreprocessor:
```
pbadillatorrealba: All base class that are configurable objects should extend ConfigObject. Also, as abstract class, it should extend ABC or have ABCMeta as abstract class

```suggestion
from abc import ABCMeta, abstractmethod
from DashAI.back.config_object import ConfigObject


class BasePreprocessor(ConfigObject, metaclass=ABCMeta):
```

DashAI/back/preprocessor/base_preprocessor.py
```diff
@@ -0,0 +1,24 @@
...
+    """Base class for DashAI preprocessor."""
+
+    TYPE: Final[str] = "Preprocessor"
+
+    @abstractmethod
+    def process(self, dataset):
```
pbadillatorrealba: I trying to enfoce that every function and method use a typing notation (although is not always possible).
Here, it could be (I don't know exactly, so check it in runtime for example (with `type`)).

We are also trying to use scikitlearn's conventions for naming methods. That is, fit to fit (whatever that is), transform for any data transformation, and predict to apply the predictor.

```suggestion
    def transform(self, dataset: DashAIDataset) -> DashAIDataset:
```

We also

DashAI/back/preprocessor/remove_column_preprocessor.py
```diff
@@ -0,0 +1,31 @@
+from base_preprocessor import BasePreprocessor
+from datasets import DatasetDict
+
+
+class RemoveColumnsPreprocessor(BasePreprocessor):
+    """Preprocessor that removes columns from the dataset."""
+
+    def process(self, dataset: DatasetDict, initial_index: int, final_index: int):
```
pbadillatorrealba: Some notes:

1. You are not respecting the original signature of the method that you extended in the base class (this could lead to errors when we check the types with for example mypy. I think that the base method should also have these parameters.
2. In this case you only allow the transformation in a range of columns. What if I wanted to use a named column? Or if I wanted separate intervals?
Perhaps the simple solution would be to have all transformations execute only one modification and then stack these. So you could have, as you said, a ColumnTransformer where you specify the transformation for each column.


```suggestion
    def transform(self, dataset: DatasetDict, col_name: str) -> DashAIDataset:
    # or col_names: List[str]
```

pbadillatorrealba (changes requested):
I think the implementation is going well, I leave you some notes.

pbadillatorrealba:
Also see the CI checks, a test is failing. 
All checks should pass to merge to dev branch.

DashAI/back/preprocessor/remove_column_preprocessor.py
Lucas-CE: One issue I've considered regarding this is that, for instance, to rename a column, I need the original index or name of the column and the new name. Therefore, using the same signature wouldn't be possible. I thought that perhaps this could be addressed by passing this information as a parameter to the object.

pbadillatorrealba:
The implementation is nice and it is almost ready.
There are two things missing to accept the pr:

1. To maintain the fit-transform interface, in each class implement a empty fit method which returns self.
2. Use the [beartype](https://github.com/beartype/beartype) decorator over the init, fit and transform methods to enforce its types.
3. Implement some tests cases for the new classes and check if they work ok.

cristian-tamblay:
The implementation is in a good state now. Try merging with staging to fix the pytest check.

DashAI/back/preprocessor/base_converter.py
```diff
@@ -0,0 +1,41 @@
...
+from abc import ABCMeta, abstractmethod
+from typing import Final
+
+from datasets import DatasetDict
+from sklearn.base import BaseEstimator, TransformerMixin
+
+from DashAI.back.config_object import ConfigObject
+
+
+class BaseConverter(ConfigObject, BaseEstimator, TransformerMixin, metaclass=ABCMeta):
```
pbadillatorrealba: So it was decided to use converter instead of preprocessor?

Why use converter and not preprocessor?

DashAI/back/preprocessor/base_converter.py
```diff
@@ -0,0 +1,41 @@
...
+    """Base class for all converters"""
+
+    TYPE: Final[str] = "Converter"
+
+    @abstractmethod
+    def fit(self, dataset: DatasetDict) -> BaseConverter:
```
pbadillatorrealba: I'm pretty sure this annotation doesn't work, since BaseConverter is not yet defined inside this scope (you are just defining it outside).

The solution to this is as follows:

```suggestion
    def fit(self, dataset: DatasetDict) -> "BaseConverter":
```

DashAI/back/preprocessor/column_dropper_by_index.py
```diff
@@ -0,0 +1,55 @@
...
+        Parameters
+        ----------
+        columns : tuple[int, int]
+            Columns to be dropped. The tuple contains the start and end index of the
+            columns to be dropped.
+        """
+        self.columns = columns
+
+    @beartype
+    def fit(self, dataset: DatasetDict) -> BaseConverter:
```
pbadillatorrealba: It seems to me that the correct thing to do here would be to return Type["BaseConverter"] (type should be imported from typing).

```suggestion
    def fit(self, dataset: DatasetDict) -> Type["BaseConverter"]:
```

DashAI/back/preprocessor/column_dropper_by_index.py
```diff
@@ -0,0 +1,55 @@
...
+
+from DashAI.back.dataloaders.classes.dashai_dataset import DashAIDataset
+from DashAI.back.preprocessor.base_converter import BaseConverter
+
+
+class ColumnDropperByIndex(BaseConverter):
+    """Converter to drop columns from the dataset by column index"""
+
+    @beartype
+    def __init__(self, columns: tuple[int, int]):
```
pbadillatorrealba: Why not call the parameter in a more descriptive way, such as `columns_index`?
```suggestion
    def __init__(self, columns_index: tuple[int, int]):
```

DashAI/back/preprocessor/column_dropper_by_name.py
```diff
@@ -0,0 +1,54 @@
...
+
+from DashAI.back.dataloaders.classes.dashai_dataset import DashAIDataset
+from DashAI.back.preprocessor.base_converter import BaseConverter
+
+
+class ColumnDropperByName(BaseConverter):
+    """Converter to drop columns from the dataset"""
+
+    @beartype
+    def __init__(self, columns: list[str] | str):
```
pbadillatorrealba: Why not call the parameter in a more descriptive way, such as `column_names`?

```suggestion
    def __init__(self, column_names: list[str] | str):
```

tests/back/preprocessor/test_drop_column_transformers.py
```diff
@@ -0,0 +1,84 @@
...
+        outputs_columns=["Species"],
+    )
+
+    return datasetdict
+
+
+def test_remove_input_column_with_column_name(
+    iris_dataset: DatasetDict, iris_dataset_petal_width_dropped: DatasetDict
+):
+    dropper = ColumnDropperByName("PetalWidthCm")
```
pbadillatorrealba: Try to use named parameters.

```suggestion
    dropper = ColumnDropperByName(column_names = "PetalWidthCm")
```

pbadillatorrealba (changes requested):
I left some other comments about styles and names. 
In addition, tests are also missing for the case of deleting columns by indexes and deleting a list of columns.

I think the tests could be done using the same test dataset.

DashAI/back/preprocessor/base_converter.py
Lucas-CE: At first there was talk of using the name preprocessor, but in reality this process is more than preprocessing the data before training. These objects are more general, they must modify the dataset, even pre-exploration.

Later, they were called Transformers, but they could be confused due to the multiple uses that the word transformers has in the area of ML, so converters were finally chosen.

pbadillatorrealba:
Almost everything is ready!

It remains to test the converters with:
- a list of columns
- a tuple indicating the start and end index of the columns to be deleted.

pbadillatorrealba (commented):
Almost ready, 

Question: What are those arrow files? They should'n be in the commit. Also, why they generates?

pbadillatorrealba (approved):

PR #156, opened 2023-09-25
POC/Structure core components in a hierarchical dependencies

pbadillatorrealba:
# Summary

This proposal proposes a new way of structuring the core components in a hierarchical way through the use of settings and imports.

Imports can be thought of as a directed acyclic graph (DAG). The DAG produces circular imports if it has a circular dependency.

To centralize the settings and customize their behavior at application startup, the following refactoring and initialization is proposed:

image

## Type of change

- Proof of concept.

## Changes

- Set an import order in `__main__`DashAI file.
- Use `dev-mode` option to configure a debug flag in the app settings (check `back/core/config.py`)
- Move app to `core/app` and and implement a factory pattern to initialize it.
- Extract the component registry from `core/config` to `core/component_registry.py` and implement a factory pattern to initialize it.
- Extract the job queue from `core/config` to `core/job_queue.py`. The factory pattern of this component will be implemented later.

## How to Test

```bash
$ pytest .
```

## Notes

Depends of #154


pbadillatorrealba:
The idea was discarded in favor of injecting dependencies.

PR #157, opened 2023-09-26
Random forest and Ada boost classifier

MartinParedesR:
# Summary

1. Added ada_boost_classifier class with its parameters.
2. Added random_forest_regressor class.

## Type of change

- Back end new feature.

## How to Test

The tests are not ready yet.

### DIFF
--- a/DashAI/back/models/scikit_learn/random_forest_regressor.py
+++ b/DashAI/back/models/scikit_learn/random_forest_regressor.py
@@ -1,0 +1,10 @@
+from sklearn.ensemble import RandomForestRegressor as _RandomForestRegressor
+
+from DashAI.back.models.scikit_learn.sklearn_like_model import SklearnLikeModel
+from DashAI.back.models.tabular_classification_model import TabularClassificationModel
+
+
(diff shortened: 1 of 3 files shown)


DashAI/back/models/scikit_learn/random_forest_regressor.py
cristian-tamblay, line 1: I think regressors aren't supported by SklearnLikeModel because the prediction uses predict_proba, not implemented in Sklearn's Regressors

cristian-tamblay (changes requested):
AdaBoost is correctly implemented, but i think we should delete the regrssor because its not compatible

pbadillatorrealba:
I think this should be corrected by making perhaps two wrappers, one for classification and one for regression. But that is for another pr.

pbadillatorrealba:
@MartinParedesR , please also check that the classifier you implemented is working correctly by executing a task on the front end.

pbadillatorrealba:
@MartinParedesR , what is the status of this change?

cristian-tamblay:
Stale, closing

PR #160, opened 2023-11-15
Configure dev front server

pbadillatorrealba:
# Summary

This pr correctly configures the back and front end of the application to use the front end development server.

## Type of change

- Front end new feature.
- Back end new feature.

## Changes

- Configured cors in FastAPI to allow calls from any URL.
- The environment variable REACT_APP_API_URL=http://localhost:8000/api was set so that the development server goes to that address when making API calls.

## How to Test

To develop, two servers must now be started (each in its own console).

- On the one hand, the front end, using 

```bash
$ cd DashAI/front && yarn start
```

and on the other, the back one, using:

```
$ python -m DashAI
```

### DIFF
--- a/DashAI/back/main.py
+++ b/DashAI/back/main.py
@@ -1,6 +1,7 @@
 import os
 
 from fastapi import FastAPI
+from fastapi.middleware.cors import CORSMiddleware
 from fastapi.responses import RedirectResponse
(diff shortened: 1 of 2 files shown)


cristian-tamblay (approved):
Nice! This helps a lot

PR #161, opened 2023-11-21
Dataset Summary Step

MillarayVVS:
# Summary

This PR includes the changes to add a new step in the New Dataset flow that allows users to see a summary of the dataset and change the column types by default after de data loading.

## Type of change

- Front end new feature.
- Back end new feature.

## Changes

- Adds the new step in front.
- Implements methods in DashAIDataset class to get and update column types.
- Implements changes on dataset endoint to get and update column types.
- Fix and adds new tests.

## How to Test

To see interface changes:
1. Go to DashAI/ and run ```python -m DashAI```
2. Open a new terminal, go to DashAI/DashAI/front and run ```yarn start```
4. Check the view in Datasets after loading a new dataset.

## Screenshots

image


DashAI/front/src/components/datasets/DatasetModal.jsx
```diff
@@ -62,12 +68,37 @@ function DatasetModal({ open, setOpen, updateDatasets }) {
...
+  const handleUpdateColumnType = async () => {
+    try {
+      await updateDatasetRequest(
+        uploadedDataset.id,
+        undefined,
+        updateColumnTypes,
+      );
+    } catch (error) {
+      enqueueSnackbar(
+        "Error while trying to update the column and data types.",
```
pbadillatorrealba, line 89: If this message is the last message in the process, then it should simply say that the upload process failed. 

Perhaps if it is for a specific reason, a description could be added to it, but then, you don't know if the reason for failure is updating the column types.

DashAI/front/src/components/datasets/DatasetModal.jsx
```diff
@@ -194,7 +236,7 @@ function DatasetModal({ open, setOpen, updateDatasets }) {
           </Button>
           <Button
             onClick={() => {
-              if (activeStep === 2) {
+              if (activeStep === 1) {
                 setReadyToUpload(true);
               }
```
pbadillatorrealba: Why this change?

DashAI/front/src/components/datasets/DatasetModal.jsx
```diff
@@ -44,6 +46,10 @@ function DatasetModal({ open, setOpen, updateDatasets }) {
   const [nextEnabled, setNextEnabled] = useState(false);
   const [newDataset, setNewDataset] = useState(defaultNewDataset);
   const [readyToUpload, setReadyToUpload] = useState(false);
+  const [uploaded, setUploaded] = useState(false);
+  const [requestError, setRequestError] = useState(false);
+  const [uploadedDataset, setUploadedDataset] = useState([]);
+  const [updateColumnTypes, setUpdateColumnTypes] = useState({});
```
pbadillatorrealba: What do you think about calling this variable `columnTypesToUpdate` or just `columnTypes`?

DashAI/front/src/components/datasets/DatasetSummaryStep.jsx
```diff
@@ -0,0 +1,169 @@
...
+      const types = await getDatasetTypesRequest(uploadedDataset.id);
+      const rowsArray = Object.keys(dataset).map((name) => {
+        return {
+          id: uuid(),
+          columnName: name,
+          example: dataset[name][0],
+          columnType: types[name].type,
+          dataType: types[name].dtype,
+        };
+      });
```
pbadillatorrealba: Why not to use an enumeration?

```suggestion
      const rowsArray = Object.keys(dataset).map((name, idx) => {
        return {
          id: idx,
          columnName: name,
          example: dataset[name][0],
          columnType: types[name].type,
          dataType: types[name].dtype,
        };
      });
```

DashAI/front/src/components/datasets/DatasetSummaryStep.jsx
```diff
@@ -0,0 +1,169 @@
...
+    apiRef.current.stopCellEditMode({ id, field });
+
+    setRows((prevRows) =>
+      prevRows.map((row) =>
+        row.id === id ? { ...row, [field]: newValue } : row,
+      ),
+    );
+
+    const columnName = rows.find((row) => row.id === id)?.columnName;
+    const updateColumns = { ...updateColumnTypes };
```
pbadillatorrealba: I believe that this operation is not doing anything and could eventually be deleted.

DashAI/front/src/api/datasets.ts
```diff
@@ -12,25 +12,33 @@ export const getDatasets = async (): Promise<IDataset[]> => {
...
+  const response = await api.get<object>(`${datasetEndpoint}/types/${id}`);
+  return response.data;
+};
 
 export const updateDataset = async (
   id: number,
-  name: string | undefined = undefined,
-  taskName: string | undefined = undefined,
+  datasetName: string | undefined = undefined,
+  columnsType:
```
pbadillatorrealba: This produces a certain ambiguity of the concepts. How about renaming it to `columnSpec`, which could encapsulate the idea of type and dtype?

DashAI/front/src/api/datasets.ts
```diff
@@ -12,25 +12,33 @@ export const getDatasets = async (): Promise<IDataset[]> => {
...
-  if (name !== undefined) {
-    params = { ...params, name };
+  if (datasetName !== undefined) {
+    formData = { ...formData, name: datasetName };
   }
 
-  if (taskName !== undefined) {
-    params = { ...params, task_name: taskName };
+  if (columnsType !== undefined) {
+    formData = { ...formData, columns: columnsType };
```
pbadillatorrealba: I believe that this operation of patching the missing attributes should not be done here, but in the component. 

I say this thinking in the idea of separation of responsibilities: the api code should only be related to the communication between the back and the front, not contain logic. For this reason, I think that the best thing to do is to pass this logic to the component that is making this call.

DashAI/back/api/api_v1/endpoints/datasets.py
```diff
@@ -77,6 +85,70 @@ async def get_dataset(dataset_id: int, db: Session = Depends(get_db)):
...
+    """
+    try:
+        file_path = db.get(Dataset, dataset_id).file_path
+        dataset: DashAIDataset = load_dataset(f"{file_path}/dataset")
+        sample = dataset["train"].sample(n=10)
+        if not file_path:
+            raise HTTPException(
+                status_code=status.HTTP_404_NOT_FOUND,
+                detail="Dataset not found",
+            )
```
pbadillatorrealba: The  `if not file_path`  condition must be just after the line 103. 
Otherwise, if the file does not exists, `load_dataset` could raise an exception.

```suggestion
        file_path = db.get(Dataset, dataset_id).file_path
        if not file_path:
            raise HTTPException(
                status_code=status.HTTP_404_NOT_FOUND,
                detail="Dataset not found",
            )
        dataset: DashAIDataset = load_dataset(f"{file_path}/dataset")
        sample = dataset["train"].sample(n=10)

```

DashAI/back/api/api_v1/endpoints/datasets.py
```diff
@@ -77,6 +85,70 @@ async def get_dataset(dataset_id: int, db: Session = Depends(get_db)):
...
+    except exc.SQLAlchemyError as e:
+        log.exception(e)
+        raise HTTPException(
+            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+            detail="Internal database error",
+        ) from e
+
+    return sample
```
pbadillatorrealba, line 118: If this is a dataframe, how fastapi is serializing it? It does the to_dict automatically?

DashAI/back/api/api_v1/endpoints/datasets.py
```diff
@@ -77,6 +85,70 @@ async def get_dataset(dataset_id: int, db: Session = Depends(get_db)):
...
+    dataset_id : int
+        id of the dataset to query.
+
+    Returns
+    -------
+    Dict
+        Dict containing column names and types.
+    """
+    try:
+        file_path = db.get(Dataset, dataset_id).file_path
```
pbadillatorrealba, line 167: I think there should also be a check here as to whether or not the dataset exists.
```suggestion
        file_path = db.get(Dataset, dataset_id).file_path
        if not file_path:
            raise HTTPException(
                status_code=status.HTTP_404_NOT_FOUND,
                detail="Dataset not found",
            )
```

DashAI/back/api/api_v1/endpoints/datasets.py
```diff
@@ -77,6 +85,70 @@ async def get_dataset(dataset_id: int, db: Session = Depends(get_db)):
...
+        column_types = get_column_types(f"{file_path}/dataset")
+        if not column_types:
+            raise HTTPException(
+                status_code=status.HTTP_404_NOT_FOUND,
+                detail="Dataset not found",
```
pbadillatorrealba, line 171: In this case, the error message should indicate the current error:

```suggestion
                detail="Column types not found", # or error while loading column types.
```

DashAI/back/api/api_v1/schemas/datasets_params.py
```diff
@@ -18,10 +18,23 @@ class SplitParams(BaseModel):
...
-    task_name: str
     dataloader: str
     dataset_name: str
     outputs_columns: List[str] = []
     splits_in_folders: bool = False
     splits: SplitParams
     dataloader_params: DataLoaderParams
+
+
+class ColumnUpdateItemParams(BaseModel):
```
pbadillatorrealba: Same consideration as before in the front end, maybe we could rename this to "ColumnSpec".

DashAI/back/dataloaders/classes/dashai_dataset.py
```diff
@@ -311,3 +311,46 @@ def save_dataset(datasetdict: DatasetDict, path: str) -> None:
...
+    dataset_features = dataset["train"].features
+    column_types = {}
+    for column in dataset_features:
+        if dataset_features[column]._type == "Value":
+            column_types[column] = {
+                "type": "Value",
+                "dtype": dataset_features[column].dtype,
+            }
+        elif dataset_features[column]._type == "ClassLabel":
+            column_types[column] = {"type": "Classlabel", "dtype": ""}
```
pbadillatorrealba: Just styles, add a final comma to the dict to make it multiline.
```suggestion
            column_types[column] = {"type": "Classlabel", "dtype": "",}
```

DashAI/back/dataloaders/classes/dashai_dataset.py
```diff
@@ -311,3 +311,46 @@ def save_dataset(datasetdict: DatasetDict, path: str) -> None:
...
+    return column_types
+
+
+@beartype
+def update_column_types(dataset_path: str, columns: Dict) -> DatasetDict:
```
pbadillatorrealba: I see this method as very complex.

I think it would be good to:

1. you add a documentation for the method.
2. Indicate how you make the change of types, if you have to load the datasets completely or just load the metadata and modify this.

tests/back/dataloaders/test_dashai_dataset.py
```diff
@@ -239,3 +244,32 @@ def test_save_to_disk_and_load():
...
+    modify_data = {
+        "SepalLengthCm": ColumnUpdateItemParams(type="Value", dtype="string"),
+        "SepalWidthCm": ColumnUpdateItemParams(type="Value", dtype="float64"),
+        "PetalLengthCm": ColumnUpdateItemParams(type="Value", dtype="string"),
+        "PetalWidthCm": ColumnUpdateItemParams(type="Value", dtype="float64"),
+        "Species": ColumnUpdateItemParams(type="Value", dtype="string"),
+    }
+
+    save_dataset(dataset, "tests/back/dataloaders/dashaidataset")
+    dataset_update = update_column_types(
```
pbadillatorrealba: Why not call this variable `updated_dataset`?

tests/back/dataloaders/test_dashai_dataset.py
```diff
@@ -239,3 +244,32 @@ def test_save_to_disk_and_load():
...
+        "tests/back/dataloaders/dashaidataset", modify_data
+    )
+    shutil.rmtree("tests/back/dataloaders/dashaidataset", ignore_errors=True)
+
+    new_features = dataset_update["train"].features
```
pbadillatorrealba: Also, check if the dict does not have any more (or less) columns than the expected:

```suggestion
    assert list(new_features.keys()) == ["SepalLengthCm", "SepalWidthCm", ["PetalLengthCm",  "PetalWidthCm", "Species"]
    new_features = dataset_update["train"].features
```

tests/back/dataloaders/test_dashai_dataset.py
```diff
@@ -239,3 +244,32 @@ def test_save_to_disk_and_load():
     assert dashai_datasetdict["train"].outputs_columns == outputs_columns
     assert dashai_datasetdict["validation"].outputs_columns == outputs_columns
     assert dashai_datasetdict["test"].outputs_columns == outputs_columns
+
+
+def test_update_column_types():
```
pbadillatorrealba: Looking at this example, I had some doubts about the update_column_types logic:

- Is it possible that the user tries to enter some unsupported type (or that this is done directly using the API)? In this case, this operation should fail or not? And if it fails, how is the error handled?
I think you should implement a new test on this (not using the API, but passing a dictionary of unsupported types to the dataset to see what happens).
- I see that in this test you are leaving the dataset without class, is this ok? If so, then you should have a test with classes and one without classes, to verify both behaviors.
- Can a dataset have more than one class column? If so, you should also test that and prove that they are all classes.
- Finally, can some of these values be null? If so, you would have to implement tests to test them. If not, test anyway to see if any exception is raised.

tests/back/api/test_dataset_api.py
```diff
@@ -92,17 +88,25 @@ def test_get_wrong_dataset(client: TestClient):
...
 
 
-def test_modify_dataset(client: TestClient):
+def test_get_types(client: TestClient):
+    response = client.get("/api/v1/dataset/types/2")
+    data = response.json()
+    assert data["SepalLengthCm"]["type"] == "Value"
+    assert data["SepalLengthCm"]["dtype"] == "float64"
+    assert data["Species"]["type"] == "Value"
+    assert data["Species"]["dtype"] == "string"
```
pbadillatorrealba: Here I would better use the whole answer as a dictionary.
pytest has a good diff engine to show that it is different in the actual and expected values :).

```suggestion
    assert data == {
        "SepalLenghtCm" {"type": "Value", "dtype": "float64"}, 
        "SepalWidth": ...}
```

tests/back/api/test_dataset_api.py
```diff
@@ -92,17 +88,25 @@ def test_get_wrong_dataset(client: TestClient):
...
+
+
+def test_modify_dataset_name(client: TestClient):
     response = client.patch(
         "/api/v1/dataset/2",
-        params={"name": "test_modify_name", "task_name": "UnknownTask"},
+        data=json.dumps({"name": "test_modify_name"}),
```
pbadillatorrealba: Why to use here json.dumps?

pbadillatorrealba (changes requested):
The new implementations are very good. 
I only left minor comments on variable names, tests, documentation and others.

DashAI/back/api/api_v1/schemas/datasets_params.py
MillarayVVS: Thank you for your recommendation, I changed the name in the front and back, I also renamed some methods to have consistency:
```update_column_types``` to ```update_columns_spec```
```get_column_types``` to ```get_columns_spec```

File: tests/back/api/test_dataset_api.py (hunk 19 above)
Comment (MillarayVVS) [reply tests/back/api/test_dataset_api.py]: Since we are using pydantic models to valite the parameters is convenient to send a JSON to the end-point, however I changed it to use json parameter instead of the data one to clean up the code.
```suggestion
        json={"name": "test_modify_name"},
```

tests/back/dataloaders/test_dashai_dataset.py
MillarayVVS: Hi! I pushed a recent commit adding some tests to the method:
- Testing errors when there is an unsoported casting such as changing a column from 'string' to 'bool'
- Testing the cast when theres only 1 class column
- Testing the cast when theres 2 class columns

When I run ```pytest tests/``` there's still one dashaidataset folder that persist after removing, could you check this changes please?

DashAI/back/api/api_v1/endpoints/datasets.py
MillarayVVS: Hi! sample returns a Dictionary already, so I think is not necessary to use to_dict in this case

PR #162, opened 2023-11-28
Added Abstract Job Class. Fixed job tests to pass with the new implementation

cristian-tamblay:
# Summary

This PR aims to make an abstraction of the runner endpoint to an Abstract Job class, named BaseJob, and then implement every kind of runnner by inheritance.
This will simplify the /job/runner endpoint to make it more general.

## Type of change

- Refactoring.

## Changes

- Implemented an abstract BaseJob class 
- Implemented RunJob class to train models
- Changed /job/runner/ POST request to /job/ POST request
- Changed frontend to match these changes
- Changed tests to match these changes
## How to Test

- You can run the normal experiment flow and train some models.
- You can run the tests


DashAI/back/api/api_v1/endpoints/jobs.py
```diff
@@ -97,35 +95,11 @@ async def enqueue_runner_job(params: JobParams, db: Session = Depends(get_db)):
...
-            detail="Internal database error",
-        ) from e
-    job = Job(
-        func=execute_run,
-        type=JobType.runner,
-        kwargs={"run_id": params.run_id, "db": db},
-    )
+    params.db = db
+    job: BaseJob = component_registry[params.job_type]["class"](**params.model_dump())
+    job.set_status_as_delivered()
```
Riul1999: This code must be surrounded by a try catch block to throw an HTTPException in case the method fails.

DashAI/back/api/api_v1/schemas/job_params.py
```diff
@@ -1,5 +1,10 @@
-from pydantic import BaseModel
+from typing import Literal
+
+from pydantic import BaseModel, ConfigDict
 
 
 class JobParams(BaseModel):
-    run_id: int
+    model_config = ConfigDict(extra="allow")
```
Riul1999, line 7: Why is it necessary to allow additional parameters or is this a good practice?

Riul1999 (commented):
The changes look very good, I leave some particular comments to review before approving the changes.

DashAI/back/api/api_v1/schemas/job_params.py
cristian-tamblay: We need to add the database as `params.db = db` so the job.run() function has access to the database.
I tried to add it as a parameter in this file as:
`db : sqlalchemy.session.Session = None` but this is not possible in pydantic because it tries to serialize a sqlalchemy database session, which is not possible

PR #163, opened 2023-11-30
remove_columns DashAIDataset method

Lucas-CE:
# Summary

The remove_columns method is implemented to DashAIDataset.

## How to Test

The test uses 2 iris datasets. One of them has all the columns, but the PetalWidth column is removed using the remove_columns method in the test. The other dataset doesnt have the PetalWidth column. 

The test attempts to prove that the 2 datasets (after removing the PetalWidth column) are equal.


tests/back/dataloaders/test_dashai_dataset.py
```diff
@@ -239,3 +243,77 @@ def test_save_to_disk_and_load():
...
+    return datasetdict
+
+
+def test_remove_columns(
+    iris_dataset: DatasetDict, iris_dataset_petal_width_dropped: DatasetDict
+):
+    assert type(iris_dataset["train"]) is DashAIDataset
+    assert type(iris_dataset_petal_width_dropped["train"]) is DashAIDataset
+    train_split: DashAIDataset = iris_dataset["train"]
+    train_dropped_split: DashAIDataset = iris_dataset["train"]
```
cristian-tamblay: These are both the same datasets. I this is the correct suggestion:
```suggestion
    train_dropped_split: DashAIDataset = iris_dataset_petal_width_dropped["train"]
```
I don't know how this test is working :rofl:

cristian-tamblay (dismissed):

cristian-tamblay (approved):

PR #164, opened 2023-12-03
xai/explainer endpoint

nataliabarca:
# Summary
Added Explainers endpoint to create, get and delete model explainers.

## Type of change
- Added `/explainers` endpoint
- (WIP) Added BaseExplainer, GlobalExplainer and LocalExplainer classes
- (WIP) Added partial dependence explainability method. 
- Created tests for the `/explainers` endpoint

## How to Test
run `pytest tests/back/api/test_explainers_api.py`


DashAI/back/api/api_v1/endpoints/explainers.py
```diff
@@ -0,0 +1,253 @@
...
+    try:
+        if run_id is not None:
+            explainers = db.scalars(
+                select(Explainer).where(Explainer.run_id == run_id)
+            ).all()
+            if not explainers:
+                raise HTTPException(
+                    status_code=status.HTTP_404_NOT_FOUND,
+                    detail="Explainer associated with Run not found",
+                )
```
pbadillatorrealba: Here's a little technical detail.

When you ask for a collection and there is no associated element, you should return a empty list and not a 404 error. Then the front end should take care of rendering that there are no elements.

In general the 404 is when you ask for a specific element and it is not found.

https://stackoverflow.com/questions/13366730/proper-rest-response-for-empty-table

DashAI/back/api/api_v1/endpoints/explainers.py
```diff
@@ -0,0 +1,253 @@
...
+        raise HTTPException(
+            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+            detail="Internal database error",
+        ) from e
+
+    return explainers
+
+
+@router.get("/{explainer_id}")
+async def get_explainer(explainer_id: int, db: Session = Depends(get_db)):
```
pbadillatorrealba:
```suggestion
async def get_explainer_by_id(explainer_id: int, db: Session = Depends(get_db)):
```

DashAI/back/api/api_v1/endpoints/explainers.py
```diff
@@ -0,0 +1,253 @@
...
+            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+            detail="Internal database error",
+        ) from e
+
+    try:
+        loaded_dataset: DashAIDataset = load_dataset(f"{dataset.file_path}/dataset")
+    except FileNotFoundError as e:
+        log.exception(e)
+        raise HTTPException(
+            status_code=status.HTTP_404_NOT_FOUND,
```
pbadillatorrealba: This error should be 500, because if the dataset exists in the database, but the associated files do not.

Then it is not a problem that the object did not exist (404), but it was due to some server error that was not saved correctly (500).

DashAI/back/api/api_v1/endpoints/explainers.py
```diff
@@ -0,0 +1,253 @@
...
+    -------
+    dict
+        Dict with the new explainer.
+
+    Raises
+    ------
+    HTTPException
+        If the explainer file cannot be saved.
+    """
+    try:
```
pbadillatorrealba, line 128: The whole function should be wrapped in this try expect, because:

1. you save the explainer in the database, which is not currently included in the current exception handler.
2. you save files, which is not considered.

With two different except, one for each case (sqlalchemyerror and oserror)

DashAI/back/api/api_v1/schemas/explainer_params.py
```diff
@@ -0,0 +1,8 @@
+from pydantic import BaseModel
+
+
+class ExplainerParams(BaseModel):
+    name: str
+    run_id: int
+    explainer_name: str
```
pbadillatorrealba: Is this the name of the explainer class?

DashAI/back/explainability/base_explainer.py
```diff
@@ -0,0 +1,41 @@
...
+logging.basicConfig(level=logging.DEBUG)
+log = logging.getLogger(__name__)
+
+
+class BaseExplainer(ConfigObject, ABC):
+    """_summary_
+
+    Args:
+        ABC (_type_): _description_
+    """
```
pbadillatorrealba, line 19: Missing docs! 👀

DashAI/back/explainability/explainers/partial_dependence.py
```diff
@@ -0,0 +1,65 @@
...
+        """
+
+        """Assumptions:
+        1. En la interfaz se podrán seleccionar los features categóricos
+        2. Se calculará para todos los features bajo los mismos parámetros
+        configurables
+
+        Cosas a considerar:
+        1. Interacting features: only continuos pairs
+        2.Centered case"""
```
pbadillatorrealba, line 43: The dosctring has yet to be finished and translated to English!

DashAI/back/explainability/explainers/partial_dependence.py
```diff
@@ -0,0 +1,65 @@
...
+        grid_resolution: int = 100,
+    ):
+        self.percentiles = percentiles
+        self.grid_resolution = grid_resolution
+
+    def explain(
+        self,
+        model: BaseModel,
+        x: DashAIDataset,
+        categorical_features,
```
pbadillatorrealba, line 26: ?
```suggestion
        categorical_features: List[str]
```

DashAI/back/explainability/explainers/partial_dependence.py
```diff
@@ -0,0 +1,65 @@
...
+class PartialDependence(GlobalExplainer):
+    COMPATIBLE_COMPONENTS = ["TabularClassificationTask"]
+
+    def __init__(
+        self,
+        percentiles: List[float],
+        grid_resolution: int = 100,
+    ):
+        self.percentiles = percentiles
+        self.grid_resolution = grid_resolution
```
pbadillatorrealba, line 20: These parameters should be declared in the schema, not directly as arguments!

DashAI/back/explainability/global_explainer.py
```diff
@@ -0,0 +1,16 @@
+from abc import abstractmethod
+
+from DashAI.back.dataloaders.classes.dashai_dataset import DashAIDataset
+from DashAI.back.explainability.base_explainer import BaseExplainer
+
+
+class GlobalExplainer(BaseExplainer):
```
pbadillatorrealba, line 7: Maybe a small docstring commenting on what is a global explainer and the explain method?

DashAI/back/explainability/local_explainer.py
```diff
@@ -0,0 +1,12 @@
+from DashAI.back.dataloaders.classes.dashai_dataset import DashAIDataset
+
+
+class LocalExplainer:
```
pbadillatorrealba, line 4: Same here, it is missing a short docstring.

DashAI/back/explainability/local_explainer.py
```diff
@@ -0,0 +1,12 @@
...
+    def __init__(self) -> None:
+        pass
+
+    def fit(self, x: DashAIDataset):
+        return self
+
+    def explain_instance(self, x: DashAIDataset):
```
pbadillatorrealba, line 11: Perhaps here for semantics, would it not be better to call x as sample (perhaps to distinguish it from the x, which is commonly used as training dataset)?

DashAI/back/explainability/local_explainer.py
```diff
@@ -0,0 +1,12 @@
+from DashAI.back.dataloaders.classes.dashai_dataset import DashAIDataset
+
+
+class LocalExplainer:
+    def __init__(self) -> None:
+        pass
+
+    def fit(self, x: DashAIDataset):
```
pbadillatorrealba, line 8: Wouldn't it be better to call it x_train?
```suggestion
    def fit(self, x_train: DashAIDataset):
```

DashAI/back/explainability/global_explainer.py
```diff
@@ -0,0 +1,16 @@
...
+    def __init__(self) -> None:
+        pass
+
+    def fit(self, x: DashAIDataset):
+        return self
+
+    @abstractmethod
+    def explain(self, model, x: DashAIDataset):
+        pass
```
pbadillatorrealba, line 16: The first X would be X_train and the second X_val or X_test?

tests/back/api/test_explainers_api.py
```diff
@@ -0,0 +1,147 @@
...
+    assert data[0]["name"] == "test_explainer"
+    assert data[0]["run_id"] == run_id
+    assert data[0]["dataset_id"] == dataset_id
+    assert data[0]["explainer_name"] == "PartialDependence"
+    assert data[0]["explainer_path"] == os.path.join(
+        settings.USER_EXPLAINER_PATH, f"{1}.pkl"
+    )
+
+
+def test_get_wrong_explainer(client: TestClient):
```
pbadillatorrealba: I have been changing the name of the tests to unify them:

```suggestion
def test_get_not_found_explainer(client: TestClient):
```

tests/back/api/test_explainers_api.py
```diff
@@ -0,0 +1,147 @@
...
+    response = client.get("/api/v1/explainer/666")
+    assert response.status_code == 404, response.text
+    assert response.text == '{"detail":"Explainer not found"}'
+
+
+def test_delete_explainer(client: TestClient):
+    response = client.delete("/api/v1/explainer/1")
+    assert response.status_code == 204, response.text
```
pbadillatorrealba, line 222: Is there any case where an explainer cannot be removed?

For example, in a case where the explainer has associated explanations? In that case, it would be necessary to decide whether to cascade deletion or not to allow deletion until the dependent objects are deleted.

tests/back/api/test_explainers_api.py
```diff
@@ -0,0 +1,147 @@
...
+    assert data[0]["run_id"] == run_id
+    assert data[0]["dataset_id"] == dataset_id
+    assert data[0]["explainer_name"] == "PartialDependence"
+    assert data[0]["explainer_path"] == os.path.join(
+        settings.USER_EXPLAINER_PATH, f"{1}.pkl"
+    )
+
+
+def test_get_wrong_explainer(client: TestClient):
+    response = client.get("/api/v1/explainer/666")
```
pbadillatorrealba, line 215: 😆
```suggestion
    response = client.get("/api/v1/explainer/666")
```

tests/back/api/test_explainers_api.py
```diff
@@ -0,0 +1,147 @@
...
+    response = client.get("/api/v1/explainer/?run_id=1")
+    assert response.status_code == 200, response.text
+    data = response.json()
+    assert data[0]["name"] == "test_explainer"
+    assert data[0]["run_id"] == run_id
+    assert data[0]["dataset_id"] == dataset_id
+    assert data[0]["explainer_name"] == "PartialDependence"
+    assert data[0]["explainer_path"] == os.path.join(
+        settings.USER_EXPLAINER_PATH, f"{1}.pkl"
+    )
```
pbadillatorrealba: This test would separate it from the previous test into a new test called `test_get_explainers_by_run_id`.

Also, here you could add another explainer associated to the same run_id? This way you could see if it indeed returns the whole associated collection.

tests/back/api/test_explainers_api.py
```diff
@@ -0,0 +1,147 @@
...
+    response = client.get("/api/v1/explainer")
+    assert response.status_code == 200, response.text
+    data = response.json()
+    assert data[0]["name"] == "test_explainer"
+    assert data[0]["run_id"] == run_id
+    assert data[0]["dataset_id"] == dataset_id
+    assert data[0]["explainer_name"] == "PartialDependence"
+    assert data[0]["explainer_path"] == os.path.join(
+        settings.USER_EXPLAINER_PATH, f"{1}.pkl"
+    )
```
pbadillatorrealba: Here I would add a couple more explainers to see if it does indeed return all of them.

DashAI/back/api/api_v1/endpoints/explainers.py
pbadillatorrealba: Check `delete_experiment`!

pbadillatorrealba (changes requested):
Ready for review, 

In general, everything is fine!
I left you some comments related to exception handling, tests, docstrings and variable names.

PR #165, opened 2023-12-07
Back/task metadata

MillarayVVS:
# Summary

Rename schema to metada in Tasks to get columns info with get Components.

## Type of change

- Refactoring.

## Changes

- Rename schema to metadata on Tasks.
- Adds get_metadata method
- get metadata with getComponents

## How to Test

Run ```pytest tests/back/api/test_components_api.py```

PR #166, opened 2023-12-11
Front/task requirements

MillarayVVS:
# Summary

Shows seleceted task requirements in Prepare Dataset step

## Type of change

- Front end new feature.

## Changes

- Gets task requirements with get components.
- Adds alert component to show the requirements.

## How to Test

To see interface changes:

1.- Go to DashAI/ and run ```python -m DashAI```
2.- Open a new terminal, go to DashAI/DashAI/front and run ```yarn start```
3.- Check the new feature on the third step of New Experiment modal.

## Screenshots

image
## Notes

This branch is created from the last PR branch (#165), only the last two commits contains the changes of this PR.

### DIFF
--- a/DashAI/front/src/components/experiments/PrepareDatasetStep.jsx
+++ b/DashAI/front/src/components/experiments/PrepareDatasetStep.jsx
@@ -21,8 +21,15 @@ function PrepareDatasetStep({ newExp, setNewExp, setNextEnabled }) {
   const [infoLoading, setInfoLoading] = useState(true);
 
   // task requirements state
-  const [taskRequirements, setTaskRequirements] = useState({});
-  const [requirementsLoading, setRequirementsLoading] = useState(true);
+  const [taskRequirements, setTaskRequirements] = useState({
+    name: "",
+    metadata: {
+      inputs_types: [],
+      inputs_cardinality: "",
+      outputs_types: [],
+      outputs_cardinality: "",
+    },
+  });
 
   // columns index state
   const [inputColumns, setInputColumns] = useState([]);
@@ -69,7 +76,6 @@ function PrepareDatasetStep({ newExp, setNewExp, setNextEnabled }) {
   };
 
   const getTaskRequirements = async () => {
-    setRequirementsLoading(true);
     try {
       const task = await getComponentsRequest({
         selectTypes: ["Task"],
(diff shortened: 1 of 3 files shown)


cristian-tamblay:
It looks really good. I have 2 concerns:
- When you say: "Input columns must be of type ClassLabel, Value" I get a bit confused, maybe instead of a comma separating it could be an "or" like this: "Input columns must be of type ClassLabel or Value"
- Some capitalization problems, maybe something like this works better:
"The input columns must be of the types ClassLabel or Value, and they should have a cardinality of n."
Same for the output.

Code is fine. I'm worried how we will introduce something to inspect the dataset while we choose which columns are input and output. But thats a problem for the future.

DashAI/front/src/components/experiments/PrepareDatasetStep.jsx
```diff
@@ -58,7 +64,31 @@ function PrepareDatasetStep({ newExp, setNewExp, setNextEnabled }) {
         console.error("Unknown Error", error.message);
       }
     } finally {
-      setLoading(false);
+      setInfoLoading(false);
```
pbadillatorrealba, line 74: What is the advantage of having independent loadings?
I am a bit worried that they may get out of sync and some of them will be left loading infinitely.

DashAI/front/src/components/experiments/PrepareDatasetStep.jsx
```diff
@@ -81,10 +111,38 @@ function PrepareDatasetStep({ newExp, setNewExp, setNextEnabled }) {
...
+    getTaskRequirements();
   }, []);
+
+  const parseListOfStrings = (stringsList) => {
+    return stringsList.join(" or ");
+  };
   return (
     <React.Fragment>
-      {!loading ? (
+      {!requirementsLoading && (
```
pbadillatorrealba: I would leave this alert independent of loading. Maybe just show it always.

pbadillatorrealba (commented):
Looks good!

I left a few questions and comments.

DashAI/front/src/components/experiments/PrepareDatasetStep.jsx
MillarayVVS: if I leave alert independent of loading the task_requirements dictionary would be empty when rendering so (in the last commit) I left a dummy dict in task_requirements definition, with this I could delete requirements loading so we don't have to worry about two loadings getting out of sync.

PR #167, opened 2023-12-18
Dependency Injection

pbadillatorrealba:
# Summary

This pr aims to replace the import-based dependency management currently used by DashAI by a dependency injection-based management.

Dependency injection is a design pattern that consists of changing the way in which different services (API for example) require their dependencies (database connectors, component registry, queues, etc...) through the concept of "inversion of control".

The idea of inversion of control is that it is not the services that import the dependencies (originally, what instantiates the dependencies), but that the dependencies are injected to the services that require them by an external mechanism that provides them. That is, control passes from the dependencies to the injection mechanism.

In this pr, dependency injection is implemented through the [python dependency injection](https://github.com/ets-labs/python-dependency-injector) framework.

It consists of 2 main components:

1. Container: class that contains all the application dependencies and configurations.
2. Wiring: Mechanism with which the framework detects and injects the dependencies to all those functions and methods that have been marked (through the `@inject` decorator) to be injected.

Below are a couple of diagrams to exemplify how this change would look like in DashAI.

**Current status**

In the current state of the application, the configuration queue and the component registry are created in config and then imported where needed.
This makes it complicated to manage its parameters, especially when configuring the app for testing:

image

**Proposed with dependency injection**

In the new architecture, all dependencies are centralized and initialized in the container. Then, wiring is the module in charge of injecting the dependencies to each of the services with the `@inject` decorator. The `Provide` class indicates which dependency is requiring the service (for example, `Provide["component_registry"]`).

image

**Testing override**

Recall that the dependencies declared in the container are what wiring provides to the services. 
Container allows to overwrite its values at runtime, which allows to easily change the dependencies (for example the database or the folders where the datasets or runs are stored) at the moment of executing tests. This can be seen in the following diagram:

Since the container is what is provided at runtime, its values can be overwritten at runtime, which allows to easily change the database or the folders where the datasets or runs are stored.

image

## Type of change

- Back end new feature.

### DashAI changes

Main changes:

- Create `DashAI/back/container.py`, module that contains the main app Container definition. This includes wiring, configurations, database connection, component registration, queues, among others.
- Changed dependency import to dependency injection in all currently declared APIs (datasets, components, experiments, jobs, predict, runs). This includes a change in how you connect to the database, from fastAPI Depends to a context manager that initializes a session (`with session_factory() as db`) .
- All non-API services (such as model_job) were also updated to use dependency injection.

Other changes:

- Refactorized the app creation in `DashAI/back/app:create_app` factory (function that, given certains parameters, generate a new app). This function is now in charge of managing the application container.
- Moved app initialization to the `DashAI/__main__` module.
- All current dependencies were moved to `DashAI/back/dependencies`.
- All the database logic was refactored in the `DashAI/back/dependencies/database/SQLiteDatabase` class, which is used later in the container.
- Delete `config.py` from core and FastAPI deps.
- Configs are now located in `DashAI/back/config.yaml` file. This change is temporal. In the CLI pr, pydantic settings based config will be restored.
- `dependency-injector` was added to requirements.txt.
- The API that serves the front end was separated from the module in charge of generating the app and is now hosted in the api path (`DashAI/back/api/front_api.py`).

### Test changes

- The tests now save their databases and associated files in the `DashAI/tmp` folder. This is achieved through several overwrites of container parameters when configuring the server for tests (`tests/api/conftest`).
- Moved conftest to the API folder (since it was not necessary to initialize a server for all other tests).
- Several tests were changed to override some dependencies (to use dependencies pointing to databases and test paths).
- Several tests were improved to be independent of each other (it happened that some tests depended on data created in other tests, which is obviously not correct).

## How to Test

Run all tests

```bash
$ pytest .
```


DashAI/__main__.py
```diff
@@ -30,17 +24,7 @@ def main(
     if dev_mode:
         logging.info("DashAI was set to development mode.")
 
-    db = SessionLocal()
-    Base.metadata.create_all(engine)
-    timer = threading.Timer(1, open_browser)
-    timer.start()
```
cristian-tamblay: This launches the browser after running the app, its easier for the user.

DashAI/back/dependencies/database/sqlite_database.py
```diff
@@ -0,0 +1,56 @@
...
+        logger.info("Using %s as SQLite path.", db_url)
+
+        self._engine = create_engine(db_url, echo=True)
+        self._session_factory = orm.scoped_session(
+            orm.sessionmaker(
+                autocommit=False,
+                autoflush=False,
+                bind=self._engine,
+            ),
+        )
```
cristian-tamblay: We are currently using the sessionmaker with default parameters (`sessionmaker(bind=engine)`), that means autoflush=True, autocommit=False. I dont actually know if this change has any repercussion on any database flush. Maybe just stick to the current one

tests/back/api/test_runs_api.py
```diff
@@ -49,7 +53,7 @@ def test_create_run(client: TestClient, experiment_id: int):
             "description": "This is a test run",
         },
     )
-    assert response.status_code == 201, response.text
+    assert response.status_code == 201
```
cristian-tamblay, line 56: If you are looking for more detailed information for debugging when a test fails, using assert response.status_code == 200, response.text is preferable. It gives immediate insights into why the response did not meet the expected status code, which can save time during the development and debugging process

tests/back/api/test_experiments_api.py
```diff
@@ -2,111 +2,138 @@
...
 
-def test_modify_experiment_step(client: TestClient):
+def test_update_experiment_step(client: TestClient):
+    """Test that an experiment step can be updated through a patch call."""
     response = client.patch(
         "/api/v1/experiment/2",
         data={"params": """{"step": "STARTED"}""", "url": ""},
     )
-    assert response.status_code == 304, response.text
+    assert response.status_code == 304
```
cristian-tamblay, line 129: If you are looking for more detailed information for debugging when a test fails, using assert response.status_code == 200, response.text is preferable. It gives immediate insights into why the response did not meet the expected status code, which can save time during the development and debugging process

cristian-tamblay (changes requested):
Looks good, some minor things to discuss

DashAI/back/dependencies/database/sqlite_database.py
pbadillatorrealba: Changed!

tests/back/api/test_runs_api.py
pbadillatorrealba: Setting the error text returned by the API does not fully express the exception raised and may hide some useful message for diagnosing bugs. Therefore, I think it is best to disable them in all its occurences.

DashAI/__main__.py
pbadillatorrealba: Restored.

cristian-tamblay (approved):

PR #169, opened 2023-12-24
Restore pydantic settings config and set local_path as unique configurable path from CLI

pbadillatorrealba:
# Summary

This pull request has two main tasks:

1. Restore the default settings handling using pydantic_settings 2.
2. Unify all paths under the local_path parameter, which can be configured from the DashAI CLI.

## Type of change

- Back end new feature.
- Refactoring.

## Changes

- The settings were moved to the original `DashAI/back/config.py` pydantic-settings model. 
- Now, the settings class is called `DefaultSettings`, which expresses the fact that these are default settings and can be overridden by the CLI.
- `create_app` now accepts the `local_path` from which all other routes (db, runs, datasets, etc...) are derived (in `_generate_config_dict` function).
- Given the above change, the api conftest now does not need to do overrides since the routes are configured when a new app is created with `create_app`.

## How to Test

Execute DashAI with no path (the execution path will be setted to `~/.DashAI`.

```bash
$ python -m DashAI 
```

The initialization should create the corresponding path (check with `cd ~&& ls -a`.
Then, if you create a dataset, it should be stored in `~/.DashAI/datasets/<dataset_name>`

Execute DashAI with a custom path

```bash
 python -m DashAI --local-path ~./DashAI_param_test
```

Then, use the same operation as before

Execute the tests:

`pytest .`

## Notes

This pr depends on #167

### DIFF
--- a/DashAI/back/app.py
+++ b/DashAI/back/app.py
@@ -1,6 +1,7 @@
 """FastAPI Application module."""
 import logging
 import pathlib
+from typing import Any, Dict, Union
 
 from fastapi import FastAPI
 from fastapi.middleware.cors import CORSMiddleware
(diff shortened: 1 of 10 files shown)


DashAI/back/app.py
```diff
@@ -1,37 +1,77 @@
...
 
-def _create_path(new_path: str) -> None:
-    full_path = pathlib.Path(new_path)
-    if not full_path.is_absolute():
-        full_path = full_path.expanduser()
+def _create_path_if_not_exists(new_path: str) -> None:
+    """Create a new path if it does not exist."""
+    if not new_path.exists():
+        logging.info("Creating new path: %s.", str(new_path))
+        new_path.mkdir(parents=True)
```
cristian-tamblay, line 23: Why we need to wrap this method?
I feel like 
```suggestion
    new_path.mkdir(parents=True, exists_ok=True)
```
is the same

cristian-tamblay (approved):
Going to approve with that minor problem. Merge when ready

PR #170, opened 2023-12-24
Configure package logging using CLI option

pbadillatorrealba:
# Summary

This pr aims to provide a mechanism to configure the DashAI subpackages loggers level by using the CLI option `--logging-level`. The levels are the same as the Python logging library. By default, the set level is "INFO"

Note that --logging-level not only affects the DashAI loggers, but also the datasets (which is set to the same level as DashAI) and the SQLAlchemy (which is only activated when logging level is DEBUG).

## Type of change

- Back end new feature.

## Changes

- Add a new `logging_level` option to DashAI CLI, which acts an enum of the available logging levels.
- A new line was added in the application factory (create_app) where the level of all the subpackages logger is set according to the value passed in the CLI.
- Improve datasets module logging.
- Add and improve some docstrings.

## How to Test

Run DashAI with logger level as `DEBUG`, upload a dataset and then, delete the datasets. Check the main and datasets generated logs:

```bash
$ python -m DashAI --logging-level DEBUG
```

Run DashAI with logger level as `WARN`, upload a dataaset and then, delete the datasets. Check the generated logs. No DashAI log should be found.

```bash
$ python -m DashAI --logging-level WARN
```

Run the test:

```bas
$ pytest. 
```

## Notes

This pr depends on #169


cristian-tamblay:
Whats the status on this? Seems mergeable but still as draft (and has a pre-commit issue)

pbadillatorrealba:
@cristian-tamblay , it is updated to the last develop and ready for revision.

DashAI/__main__.py
```diff
@@ -9,21 +16,55 @@
 from DashAI.back.app import create_app
 
 
-def open_browser():
+class LoggingLevel(str, Enum):
```
Riul1999: Is it ok if we leave this definition here? I think it could be moved to DashAI core or maybe to another file at the same level as ___main__.py_.

DashAI/back/app.py
```diff
@@ -61,25 +65,65 @@ def _generate_config_dict(
...
     settings["FRONT_BUILD_PATH"] = pathlib.Path(settings["FRONT_BUILD_PATH"]).absolute()
+    settings["LOGGING_LEVEL"] = getattr(logging, logging_level)
 
     return settings
 
 
-def create_app(local_path: Union[pathlib.Path, None] = None) -> FastAPI:
+def create_app(
+    local_path: Union[pathlib.Path, None] = None,
+    logging_level: Literal[
```
Riul1999, line 76: Why do you use the value instead of the enum object?, I think it is less maintainable that way.

Riul1999 (commented):
The changes look great, I have had a lot of problems trying to debug the code leaving some log levels in INFO, so this change will be very very useful.
I think it is necessary to write a document specifying when to use each log level, and in the future we will need to refactor the code to use the log level defined when creating the app.

cristian-tamblay:
I was playing with this the other day when I needed to debug database code in another branch but the following keeps happening.
We have multiple libraries like sqlalchemy or huggingface that have their own loggers setted up, so even if we run 
`python -m DashAI --logging-level CRITICAL`, we get a bunch of crap printed:
image

And its even worse, these logging prints are not going through even if we have logging level debug activated for some reason:
```
        try:
            logger.debug("Generating a new database session.")
            yield session
        except Exception:
            logger.exception(
                "Session rollback: An exception was raised while trying to operate "
                "the database."
            )
            session.rollback()
            raise
        finally:
            logger.debug("Closing the current session.")
            session.close()
```
But if change that code to this:

```
        try:
            print("Generating a new database session.")
            yield session
        except Exception:
            print(
                "Session rollback: An exception was raised while trying to operate "
                "the database."
            )
            session.rollback()
            raise
        finally:
            print("Closing the current session.")
            session.close()
```
We suddenly can debug. I don't really know what is causing this.

DashAI/__main__.py
pbadillatorrealba: I moved the enum to core/enums/logging_levels module.

DashAI/back/app.py
pbadillatorrealba: Thus, I can use the create_app function (for example in the tests) without having to import the enum. 
In fact, I use the enum only so that I can have limited categories in the typer CLI (which is not possible otherwise).

pbadillatorrealba:
@Riul1999 , I corrected the comment of the enum in main and added a small documentation indicating how to run the app with different logging levels (and I also took the opportunity to comment the missing option, --local-path).

Note that the --logging-level not only affects the DashAI loggers, but also the datasets (which is set to the same level as DashAI) and the SQLAlchemy (which is only activated when logging level is DEBUG).

@cristian-tamblay , I tested the problem you mentioned with the DEBUG logging level (python -m DashAI --logging-level DEBUG ) and the logging you mentioned worked. I don't know if I didn't replicate the problem well?

I think I applied all the changes (except for the literal types which I think we could discuss), so it's ready to be checked again ;)

Riul1999:
> @Riul1999 , I corrected the comment of the enum in main and added a small documentation indicating how to run the app with different logging levels (and I also took the opportunity to comment the missing option, --local-path).
> 
> Note that the --logging-level not only affects the DashAI loggers, but also the datasets (which is set to the same level as DashAI) and the SQLAlchemy (which is only activated when logging level is DEBUG).
> 
> @cristian-tamblay , I tested the problem you mentioned with the DEBUG logging level (python -m DashAI --logging-level DEBUG ) and the logging you mentioned worked. I don't know if I didn't replicate the problem well?

I have checked it on my PC and the log levels work fine, I have used DEBUG, INFO and WARN.

For the libraries I found a [post](https://stackoverflow.com/questions/75552504/how-to-change-the-loglevel-or-silence-all-logs-for-a-third-party-library-in-pyth) that shows how to change the log level of the libraries one by one, I will look for a cleaner solution but I think this will work for now.

> 
> I think I applied all the changes (except for the literal types which I think we could discuss), so it's ready to be checked again ;)

cristian-tamblay (approved):

PR #171, opened 2024-01-04
In/Out-columns in DashAIDataset

MillarayVVS:
# Summary

Datasets Refactor deleting all the usage of input and output columns on other components.

## Type of change

- Refactoring.

## Changes

Remove use of input and output columns on DashAIDataset class:
- Delete getters and setters.
- Modify save and load fuctions, now they don't include input and output columns.
- Modify validation of input and output columns, now they are passed as parameters to the method.

With the modifications above it was necessary to modify all the usage of DashAIDataset on other components:

Task:
- Modify prepare and validate dataset methods, now the input and output columns are passed as parameters.

DataLoader:
- Delete input and output columns to create a new DashAIDataset.

Metrics:
- Adds parameter with output columns to use in prepare to metric function.

Model:
- Refactor of fit function, now the dataset is divided in two on Model Job so this method receive a dataset with the input columns and a dataset with the output columns as two different parameters.
- Same with predict function, now this method only receive the dataset with the output columns.

image

All this changes include the tests refactoring.

## How to Test

1.- Run ```pytest tests/```


DashAI/back/dataloaders/classes/dashai_dataset.py
```diff
@@ -249,45 +163,6 @@ def sample(
...
-    if set(inputs + outputs) != set(names):
-        raise ValueError(
-            "The union of the elements of inputs and outputs list must be equal to "
-            "elements in the list of names."
-        )
-
-
 @beartype
 def load_dataset(dataset_path: str) -> DatasetDict:
     """Load a datasetdict with dashaidatasets inside.
```
pbadillatorrealba:
```suggestion
    """Load a DashAI dataset from its path.

         This process cast each split into a DashAIdataset object.
```

DashAI/back/dataloaders/classes/dashai_dataset.py
```diff
@@ -352,6 +216,81 @@ def save_dataset(datasetdict: DatasetDict, path: str) -> None:
...
 
 
+@beartype
+def validate_inputs_outputs(
+    datasetdict: DatasetDict,
+    inputs: List[str],
+    outputs: List[str],
+) -> None:
+    """Validate the columns to be chosen as input and output.
+    The algorithm considers those that already exist in the dataset.
```
pbadillatorrealba, line 227: Add a new line between the first line and the next paragraph in a docstring.

DashAI/back/dataloaders/classes/dashai_dataset.py
```diff
@@ -352,6 +216,81 @@ def save_dataset(datasetdict: DatasetDict, path: str) -> None:
...
+            "Inputs and outputs cannot have more elements than names. "
+            f"Number of inputs: {len(inputs)}, "
+            f"number of outputs: {len(outputs)}, "
+            f"number of names: {len(dataset_features)}. "
+        )
+        # Validate that inputs and outputs only contain elements that exist in names
+    if not set(dataset_features).issuperset(set(inputs + outputs)):
+        raise ValueError(
+            "Inputs and outputs can only contain elements that exist in names."
+        )
```
pbadillatorrealba: I think it is a good idea to show the extra features detected:
```suggestion
        raise ValueError(
            "Inputs and outputs can only contain elements that exist in names. Extra elements: %s",
            str(set(input+output). difference(set(datasets_features))) # check this op
        )
```

DashAI/back/dataloaders/classes/dashai_dataset.py
```diff
@@ -352,6 +216,81 @@ def save_dataset(datasetdict: DatasetDict, path: str) -> None:
         )
 
 
+@beartype
+def validate_inputs_outputs(
+    datasetdict: DatasetDict,
+    inputs: List[str],
+    outputs: List[str],
+) -> None:
+    """Validate the columns to be chosen as input and output.
```
pbadillatorrealba, line 226: Nothing happens if the lists are empty?

DashAI/back/dataloaders/classes/dashai_dataset.py
```diff
@@ -352,6 +216,81 @@ def save_dataset(datasetdict: DatasetDict, path: str) -> None:
...
+        )
+        # Validate that inputs and outputs only contain elements that exist in names
+    if not set(dataset_features).issuperset(set(inputs + outputs)):
+        raise ValueError(
+            "Inputs and outputs can only contain elements that exist in names."
+        )
+
+
+@beartype
+def parse_columns_indices(dataset_path: str, indices: List[int]) -> List[str]:
```
pbadillatorrealba, line 260: Add documentation.

DashAI/back/dataloaders/classes/dashai_dataset.py
```diff
@@ -352,6 +216,81 @@ def save_dataset(datasetdict: DatasetDict, path: str) -> None:
...
+    dataset = load_dataset(dataset_path=dataset_path)
+    dataset_features = list((dataset["train"].features).keys())
+    names_list = []
+    for index in indices:
```
pbadillatorrealba, line 273: `indices` comes from the front?
Maybe its a good idea to preprocess it outside this function, maybe in the same request code in the API.

DashAI/back/dataloaders/classes/dashai_dataset.py
```diff
@@ -352,6 +216,81 @@ def save_dataset(datasetdict: DatasetDict, path: str) -> None:
...
+        if index <= len(dataset_features):
```
pbadillatorrealba: If the validation does not fail, this should not happen right?

Anyway, I suggest adding an else where you throw an exception if this case occurs.

DashAI/back/dataloaders/classes/dashai_dataset.py
```diff
@@ -352,6 +216,81 @@ def save_dataset(datasetdict: DatasetDict, path: str) -> None:
...
+    output_columns : List[str]
+        List with the output columns labels
+
+    Returns
+    -------
+    Dict
+        Dict with the splits divided in x and y tuple
+    """
+    input_columns_dataset = copy.deepcopy(dataset)
+    output_columns_dataset = copy.deepcopy(dataset)
```
pbadillatorrealba: I think it would be a good idea to see if the original dataset object implements copy and use that.
It could be that, making an efficient memory management, this object does not contain the dataset, but only a reference to it and that at the moment of deepcopy, only the reference is copied and not the complete dataset.

DashAI/back/dataloaders/classes/dashai_dataset.py
```diff
@@ -352,6 +216,81 @@ def save_dataset(datasetdict: DatasetDict, path: str) -> None:
...
+@beartype
+def select_columns(
+    dataset: DatasetDict, input_columns: List[str], output_columns: List[str]
+) -> Tuple[DatasetDict, DatasetDict]:
+    """Load and prepare the dataset into dataframes to use in models.
+
+    Parameters
+    ----------
+    dataset : DashAIDataset
+        Dataset to format
```
pbadillatorrealba: Update the docstring.

DashAI/back/dataloaders/classes/dashai_dataset.py
pbadillatorrealba: Also, `select_colums` will not make copies?

DashAI/back/dataloaders/classes/dashai_dataset.py
```diff
@@ -352,6 +216,81 @@ def save_dataset(datasetdict: DatasetDict, path: str) -> None:
...
+    input_columns : List[str]
+        List with the input columns labels
+    output_columns : List[str]
+        List with the output columns labels
+
+    Returns
+    -------
+    Dict
+        Dict with the splits divided in x and y tuple
```
pbadillatorrealba: Update output docs

DashAI/back/api/api_v1/endpoints/predict.py
```diff
@@ -110,17 +102,19 @@ async def perform_predict(
...
-        filepath_or_buffer=input_file, temp_path=tmp_path, params={"data_key": "data"}
+        filepath_or_buffer=input_file,
+        temp_path=tmp_path,
+        params={"data_key": "data"},
     )
+    # TODO Extract this Code to DashAIDataset
     input_df = pd.DataFrame(raw_dataset["train"])
-    # TODO: Use feature_names from Experiment
-    input_df = input_df.reindex(columns=json.loads(dat.feature_names))
+    input_df = input_df.reindex(columns=exp.input_columns)
```
pbadillatorrealba, line 114: Does this functionality was not done in `select_columns`?

DashAI/back/dataloaders/classes/dashai_dataset.py
pbadillatorrealba, line 1: Since the columns are now handled externally in the experiment, does it make sense to continue building a wrapper?
Why not just use Datasets directly?

DashAI/back/dataloaders/classes/dataloader.py
```diff
@@ -92,22 +92,10 @@ def extract_files(self, dataset_path: str, file: UploadFile) -> str:
 
     def _check_split_values(
         self,
-        dataset: DatasetDict,
         train_size: float,
         test_size: float,
         val_size: float,
-        stratify: bool = False,
```
pbadillatorrealba, line 99: Will stratify no longer be implemented?

DashAI/back/models/scikit_learn/sklearn_like_model.py
```diff
@@ -22,41 +21,24 @@ def load(filename: str) -> None:
...
-        x, y = self.format_data(dataset)
-        return super().fit(x, y)
+        x_pandas = x.to_pandas()
+        y_pandas = y.to_pandas()
+        return super().fit(x_pandas, y_pandas)
 
-    def predict(self, dataset: DashAIDataset):
-        x, y = self.format_data(dataset)
-        return super().predict_proba(x)
+    def predict(self, x: DashAIDataset):
```
pbadillatorrealba: Document this function

DashAI/back/models/scikit_learn/sklearn_like_model.py
```diff
@@ -22,41 +21,24 @@ def load(filename: str) -> None:
...
         """Fit the estimator.
 
         Parameters
         ----------
-        dataset : DashAIDataset
-            The training dataset.
+        x : pd.DataFrame
+            Dataframe with the input data.
+        y : pd.DataFrame
+            Dataframe with the output data.
```
pbadillatorrealba: Why not to use `X_train` and `y_train` here , and in `X_pred` in the prediction function?

pbadillatorrealba (changes requested):
I left comments on the code of the changes, I am still pending with the review of the tests.

DashAI/back/dataloaders/classes/dashai_dataset.py
MillarayVVS: I added an Error to throw when this happens

DashAI/back/api/api_v1/endpoints/predict.py
Riul1999: No, the above code reorders the columns in the dataset to match the column order when the model was trained. The `select_columns` method of **DashAIDataset** selects a subset of the actual columns in the **Dataset**, but does not affect the column order.

DashAI/back/dataloaders/classes/dataloader.py
Riul1999: We are going to move the _split_dataset_ method from the **Dataloader** class to the **DashAIDataset** class, because the splitting of the dataset is done at experiment creation (linked to the **DashAIDataset**) instead of dataset creation (linked to the **Dataloader**). So we will add the option to _stratify_ in this new method.

pbadillatorrealba (approved):
Looks good to me!

PR #174, opened 2024-01-18
Restore DashAI init file

inunezn7:
# Summary

A __ init __ file is missing in the DashAI folder. This causes that when installing the package via "pip install" it does not build or update the distribution.

I simply renamed the  __main __ file to __ init __ to fix this quickly, but I don't know if this has an impact on any other part of the code. Please check this.

## Type of change

- Bug fix.

## Changes

- Rename file (__ main __ --> __ init __)
- Update setup.py to run main function

## How to Test

When installing DashAI using "pip install ." the package is built and installed successfully.


pbadillatorrealba:
The idea of the __main__.py is that you can run dashai through the python DashAI command.
I would suggest that we leave it as is and do an init that imports from main:

In __main__, change :

```python
if __name__ == "__main__":
    run()
```

to: 

```python
if __name__ == "__main__":
    typer.run(main)
def run():
    typer.run(main)
````

And the new __init__ file:

```python
from DashAI.__main__ import run

if __name__ == "__main__":
    run()
````

This ensures that the entrypoint still works. In fact, maybe we could in this same pr add the DashAI entrypoint (with capital letters).
I tested it locally and it worked correctly. what do you think?

pd: the branch should be updated!

inunezn7:
It works fine for me too. I have a question though. Why do we need the run function? Is not enough to have call the main function? Is it because of the typer?
Regarding the DashAI entrypoint, you mean having both entrypoints with and without capital letters?
I pushed these two changes.

pbadillatorrealba:
Because from what I understand, expressions (like "`DashAI = DashAI:typer.run(something)`" cannot be run at the entry point, it only can accept functions.

pbadillatorrealba (approved):

pbadillatorrealba (changes requested):
There are some tests that are failing

Please check the before merging.

inunezn7:
Done! All tests are passing now

pbadillatorrealba (approved):

cristian-tamblay (changes requested):
Sorry for the delay @inunezn7 . I can't merge this into develop:
- This is throwing this warning when run as `python -m DashAI`: `RuntimeWarning: 'DashAI.__main__' found in sys.modules after import of package 'DashAI', but prior to execution of 'DashAI.__main__'; this may result in unpredictable behaviour`

This makes sense because the `__init__.py` file is being imported after `__main__.py`, but `__main__.py` was already imported when calling `python -m DashAI`. This has the potential to create circular imports in the future. I don't really know what is the solution to this @pbadillatorrealba . Do you think is a good idea to merge this? We could just duplicate the code on `__init__` and `__main__` instead of importing them

Also I think after the depedency injection PR, we can no longer install DashAI via `pip install .` Have you seen this error @inunezn7 or @Lucas-CE ?

PR #176, opened 2024-01-23
Change splits in Experiment configuration

MillarayVVS:
# Summary

Adds back-end logic to support the change of splits configuration in Prepare Dataset Step.

## Type of change

- Front end new feature.
- Back end new feature.
- Refactoring.

## Changes

- Shows default splits in Prepare Dataset Step
- Save split configuration adding status to know is the splits have changed and if the new configuration is radom splits
- Use this new configuration in ModelJob flow by calling a new function update_dataset_splits
- Include tests for update_dataset_splits

## How to Test

1.- To check test run ```pytest tests/```
2.- To check correct functionality of this new feature from the front:
- Run ```python -m DashAI```
- Go to ```dashai/front``` and run ```yarn start```
- Create new experiments with different configuration of splits and verify the diffence between the results.


DashAI/back/dataloaders/classes/dashai_dataset.py
```diff
@@ -217,6 +226,169 @@ def save_dataset(datasetdict: DatasetDict, path: str) -> None:
...
+    test_mask = np.isin(np.arange(n), test_indices)
+    val_mask = np.isin(np.arange(n), val_indices)
+
+    # Get the underlying table
+    table = dataset.data
+
+    # Create separate tables for each split
+    train_table = table.filter(pa.array(train_mask))
+    test_table = table.filter(pa.array(test_mask))
+    val_table = table.filter(pa.array(val_mask))
```
pbadillatorrealba, line 364: What difference would it make to use [select](https://huggingface.co/docs/datasets/v1.8.0/package_reference/main_classes.html#datasets.Dataset.select)?
According to what I saw in the documentation, you can pass the indexes directly, making the code a little more readable.

DashAI/back/dataloaders/classes/dashai_dataset.py
```diff
@@ -217,6 +226,169 @@ def save_dataset(datasetdict: DatasetDict, path: str) -> None:
...
+
+@beartype
+def split_indices(
+    total_rows: int,
+    train_size: float,
+    test_size: float,
+    val_size: float,
+    seed: Union[int, None] = None,
+    shuffle: bool = True,
+) -> Tuple:
```
pbadillatorrealba:
```suggestion
) -> Tuple[List, List, List]:
```

I take advantage of this same space to ask, why not return the np.array instead of the lists?

I think that conversion could be avoided without problems.

DashAI/back/dataloaders/classes/dashai_dataset.py
```diff
@@ -217,6 +226,169 @@ def save_dataset(datasetdict: DatasetDict, path: str) -> None:
...
+        Proportion of the dataset for validation split (in 0-1).
+    seed : Union[int, None], optional
+        Set seed to control to enable replicability, by default None
+    shuffle : bool, optional
+        If True, the data will be shuffled when splitting the dataset,
+        by default True.
+
+    Returns
+    -------
+    Tuple
```
pbadillatorrealba: Remember to update to the final Tuple type.

DashAI/back/dataloaders/classes/dashai_dataset.py
```diff
@@ -217,6 +226,169 @@ def save_dataset(datasetdict: DatasetDict, path: str) -> None:
...
+        Train, Test and Validation indices.
```
pbadillatorrealba:
```suggestion
        Train, Test and Validation indexes.
```

DashAI/back/dataloaders/classes/dashai_dataset.py
```diff
@@ -217,6 +226,169 @@ def save_dataset(datasetdict: DatasetDict, path: str) -> None:
...
+@beartype
+def split_indices(
+    total_rows: int,
+    train_size: float,
+    test_size: float,
+    val_size: float,
+    seed: Union[int, None] = None,
+    shuffle: bool = True,
+) -> Tuple:
+    """Generate lists with train, test and validation indices.
```
pbadillatorrealba:
```suggestion
    """Generate lists with train, test and validation indexes.
```

DashAI/back/dataloaders/classes/dashai_dataset.py
```diff
@@ -403,3 +575,40 @@ def get_dataset_info(dataset_path: str) -> object:
...
+    """_summary_
+
+    Args:
+        datasetdict (DatasetDict): Dataset to update splits
+        new_splits (object): Object with the new train, test and validation config
+        is_random (bool): If the new splits are random by percentage
+
+    Returns:
+        DatasetDict: New DatasetDict with the new splits configuration
+    """
```
pbadillatorrealba: Improve documentation.

DashAI/back/api/api_v1/endpoints/datasets.py
```diff
@@ -239,14 +241,21 @@ async def upload_dataset(
         dataset = to_dashai_dataset(dataset)
 
         if not parsed_params.splits_in_folders:
-            dataset = dataloader.split_dataset(
-                dataset,
+            n = len(dataset["train"])
+            train_indices, test_indices, val_indices = split_indices(
```
pbadillatorrealba: Prefer indexes over indices:

```suggestion
            train_indexes, test_indexes, val_indexes = split_indexes(
```

DashAI/back/api/api_v1/endpoints/datasets.py
```diff
@@ -239,14 +241,21 @@ async def upload_dataset(
...
                 parsed_params.splits.test_size,
                 parsed_params.splits.val_size,
                 parsed_params.splits.seed,
                 parsed_params.splits.shuffle,
             )
+            dataset = split_dataset(
+                dataset["train"],
+                train_indices=train_indices,
+                test_indices=test_indices,
+                val_indices=val_indices,
```
pbadillatorrealba:
```suggestion
                train_indexes=train_indexes,
                test_indexes=test_indexes,
                val_indexes=val_indexes,
```

DashAI/back/dataloaders/classes/dashai_dataset.py
```diff
@@ -217,6 +226,169 @@ def save_dataset(datasetdict: DatasetDict, path: str) -> None:
...
+        )
+    if val_size < 0 or val_size > 1:
+        raise ValueError(
+            "val_size should be in the (0, 1) range "
+            f"(0 and 1 not included), got {val_size}"
+        )
+
+
+@beartype
+def split_indices(
```
pbadillatorrealba: Prefer indexes over indices

```suggestion
def split_indexes(
```

pbadillatorrealba: Same in all function params names and params.

DashAI/back/job/model_job.py
```diff
@@ -114,6 +116,16 @@ def run(self) -> None:
...
+                splits = json.loads(experiment.splits)
+                if splits["has_changed"]:
+                    new_splits = {
+                        "train": splits["train"],
+                        "test": splits["test"],
+                        "validation": splits["validation"],
+                    }
+                    loaded_dataset = update_dataset_splits(
+                        loaded_dataset, new_splits, splits["is_random"]
+                    )
```
pbadillatorrealba, line 128: When might the splits be updated?
I think that for a given experiment, it should always be the same indices.

DashAI/front/src/components/experiments/NewExperimentModal.jsx
```diff
@@ -94,6 +100,7 @@ export default function NewExperimentModal({
   };
 
   const uploadNewExperiment = async () => {
+    console.log(newExp.splits);
```
pbadillatorrealba: Delete console.log.

DashAI/front/src/components/experiments/NewExperimentModal.jsx
```diff
@@ -43,7 +43,13 @@ const defaultNewExp = {
   task_name: "",
   input_columns: [],
   output_columns: [],
-  splits: { training: 0, validation: 0, testing: 0 },
+  splits: {
+    has_changed: false,
+    is_random: true,
+    train: 0.7,
+    validation: 0.1,
+    test: 0.2,
```
pbadillatorrealba: If these are the default values, I think this ratio would be better: 

```suggestion
    train: 0.6,
    validation: 0.2,
    test: 0.2,
```

DashAI/front/src/components/experiments/PrepareDatasetStep.jsx
```diff
@@ -43,9 +43,9 @@ function PrepareDatasetStep({ newExp, setNewExp, setNextEnabled }) {
     test: [],
   };
   const defaultPartitionsPercentage = {
-    train: 60,
-    validation: 20,
-    test: 20,
+    train: 0.7,
+    validation: 0.1,
+    test: 0.2,
```
pbadillatorrealba: Same here.

DashAI/front/src/components/experiments/SplitDatasetRows.jsx
```diff
@@ -19,43 +18,35 @@ function SplitDatasetRows({
...
   const checkSplit = (train, validation, test) => {
-    return train + validation + test === 100;
+    return train + validation + test === 1;
   };
 
   const handleRowsPreferenceChange = (event) => {
     if (event.target.value === "splitByIndex") {
-      setRowsPartitionsPercentage(defaultPartitionsPercentage);
+      setIsRandom(false);
+      setRowsPartitionsPercentage({ train: 70, test: 20, validation: 10 });
```
pbadillatorrealba: Update default values here.

pbadillatorrealba (changes requested):
Looks good!

In general, I left comments regarding documentation and variable names. Just a couple of implementation issues (such as update splits).

DashAI/back/dataloaders/classes/dashai_dataset.py
MillarayVVS: Hi! I didn't code this section originally, I tried on using select but I found some problems implementing it,  Select and Filter method from Dataset give a new Dataset with the selected rows but is doesn't update the original table (so the test returns error). I think this is why we are filtering the Table not the Dataset. When I try to use the select or filter method with the Table instaed of the Dataset (passing only the indices directly as parameters without using masks) it throws IndexErrors because of indx out of bounds.

I'm currently trying to track the error as the indices I'm giving to select rows seems to be ok.

DashAI/back/dataloaders/classes/dashai_dataset.py
MillarayVVS: Hi! I did the conversion because theres other cases where split_dataset takes List and not np.array, but now I removed the conversion and instaed added the NDArray type to the parameters of split_dataset, you can check this on the recent "remove np.array to list conversion on split_indexes function" commit.

MillarayVVS: I just checked that this change throws an error using python 3.8, I don't have this issue with python 3.10

DashAI/back/job/model_job.py
MillarayVVS: This is implemented this way because the datasets come already with a separation by default, so if the train, test and validation splits aren't changed by the user the code doesn't update the splits, but if when creating a new experiment the user wants a new combination for them different to the one setted in the dataset upload, the code update the dataset splits.

This could eventually be deleted if the dataset isn't separated anymore when loading it and is only divided when configurating the experiment, so the difference between both cases will no longer exist.

DashAI/back/dataloaders/classes/dashai_dataset.py
pbadillatorrealba: What's the error?

DashAI/back/job/model_job.py
pbadillatorrealba: Ok!

DashAI/back/dataloaders/classes/dashai_dataset.py
pbadillatorrealba: In this case it is different because the dataset is supposed to be already saved on disk. 
So, if you modify the dataset in memory and save it in another file, you would not be changing the original one.

DashAI/back/dataloaders/classes/dashai_dataset.py
MillarayVVS: I can't include List and NDArray with the operand | since these are both "Generic Alias" types, you can check the error on the failed pytest check.

This is the only issue in Github I have found by the moment: https://github.com/python/cpython/issues/86399, so if we keep working with python 3.8 and 3.9 I think I will have to convert one of the cases to only accept one of the types in the function.

pbadillatorrealba: Ah ok, let's leave it at that then :)

PR #179, opened 2024-01-30
First aproach to Pydantic Schemes

Riul1999:
# Summary

Added a new way to declare and generate schemas for Configurable Components, a pydantic JSON schema.

All changes are summarized on [this board](https://miro.com/app/board/uXjVN5C8EmU=/) in the Component Schema Flow section.

## Type of change

- Refactoring.

## Changes

- **BaseSchema** class, which all schemas must inherit.
- Type field functions, designed to create custom pydantic types that help create the desired JSON schema and perform validations. 3 types are implemented:
  - _int_field_, for integer params, can set a minimum and a maximum.
  - _string_field_, for enum params, must define the posible values.
  - _component_field_, for component params, must define the parent component.
- Function _fill_objects_ to perform the transformation step, which consists of transforming all component fields into real components.
- Validation and Transformation step in **ModelJob**, now the model user input parameters are validated and transformed, so the model receives the parameters specified by the schema.
- Create test_schemas to check that the generated schemas validate and transform the input parameters and can handle nullable and union fields.
- Refactor test_predict to use dummy components instead of real components. This is to decouple the test from the other parts of DashAI.
- Move conftest.py from the api folder to the back folder to initialize the container for all tests.
- Some minor corrections:
  - Move and refactor the _get_schema_ method, it is now in the **ConfigObj** class and uses the **SCHEMA** field of the current component..
  - Remove _get_schema_ method from **BaseModel** and **BaseDataloader**.
  - Remove _get_schema_ method from dummy models in tests.

## How to Test

Run in the terminal ``pytest ./tests/back/schemas/test_schemas.py``

## Notes

The change breaks some code, especially code that uses the **SCHEMA** field of the components and the _get_schema_ method. Run all tests with ``pytest .`` to check that nothing is broken. I run it but you never know 😄.


DashAI/back/core/schema_fields/string_field.py
```diff
@@ -0,0 +1,65 @@
...
+):
+    """Function to create a pydantic-like string type.
+
+    Parameters
+    ----------
+    description: str
+        Description of the field.
+    default: str
+        The default value to show to the user.
+    enum: List[str]
```
pbadillatorrealba, line 46: Could there be any string parameter that is not a category?

DashAI/back/core/schema_fields/string_field.py
```diff
@@ -0,0 +1,65 @@
...
+    Checks if the input str is in the enum.
+
+    Parameters
+    ----------
+    enum: List[str]
+        All the posible string values of the param.
+
+    Returns
+    -------
+    str -> str
```
pbadillatorrealba:
```suggestion
    Callable[str, str]
```

DashAI/back/core/schema_fields/string_field.py
```diff
@@ -0,0 +1,65 @@
+from typing import List
+
+from pydantic import AfterValidator, Field
+from typing_extensions import Annotated
+
+
+def __check_choices(enum: List[str]):
```
pbadillatorrealba:
```suggestion
def __check_choices(enum: List[str]) -> Callable[str, str]:
```

DashAI/back/core/schema_fields/string_field.py
```diff
@@ -0,0 +1,65 @@
...
+        All the posible string values of the param.
+
+    Returns
+    -------
+    str -> str
+        A function that inspect the input string.
+    """
+
+    def check_str_in_enum(x: str) -> str:
+        assert x in enum, f"{x} is not in the enum"
```
pbadillatorrealba: Asserts should be used only for testing, not at runtime. This is because, unless you are running a testing session, they are disabled!

https://stackoverflow.com/questions/40182944/whats-the-difference-between-raise-try-and-assert

```suggestion
        if x not in enum:
            raise ValueError(f"{x} is not in the enum")
```

DashAI/back/core/schema_fields/string_field.py
```diff
@@ -0,0 +1,65 @@
...
+
+    Parameters
+    ----------
+    enum: List[str]
+        All the posible string values of the param.
+
+    Returns
+    -------
+    str -> str
+        A function that inspect the input string.
```
pbadillatorrealba:
```suggestion
        A function that checks if the string is within the possible values specified in the enum.
```

DashAI/back/core/schema_fields/component_field.py
```diff
@@ -0,0 +1,46 @@
...
+    -------
+    type[ComponentType]
+        A pydantic-like type to represent the component.
+
+    Raises
+    ------
+    ValidationError
+        If the component name of the field is not in the registry.
+    ValidationError
+        If the component name of the field is not a subclass of the parent component.
```
pbadillatorrealba: Comment in the meantime (not yet implemented :( )

DashAI/back/core/schema_fields/utils.py
```diff
@@ -0,0 +1,44 @@
+from dependency_injector.wiring import Provide, inject
+
+from DashAI.back.core.schema_fields.base_schema import BaseSchema
+
+
+@inject
+def fill_objects(
+    schema_instance: BaseSchema,
+    component_registry=Provide["component_registry"],
+) -> dict:
```
pbadillatorrealba, line 10: Is this function working?
I ask because of the dependency injection.

pbadillatorrealba (commented):
The changes look good!

I left some comments.

DashAI/back/core/schema_fields/int_field.py
```diff
@@ -0,0 +1,46 @@
...
+
+from pydantic import Field
+from typing_extensions import Annotated
+
+
+def int_field(
+    description: str,
+    default: int,
+    minimum: Optional[int] = None,
+    maximum: Optional[int] = None,
```
pbadillatorrealba: I think it would be a good idea here to keep the pydantic notation of ge, gt, le, lt:
https://docs.pydantic.dev/latest/concepts/fields/#numeric-constraints

Also, it would be an equally good idea to add the greater than and less than.


```suggestion
    ge: Optional[int] = None,
    gt: Optional[int] = None,
    le: Optional[int] = None,
    lt: Optional[int] = None,
```

DashAI/back/core/schema_fields/int_field.py
```diff
@@ -0,0 +1,46 @@
+from typing import Optional
+
+from pydantic import Field
+from typing_extensions import Annotated
+
+
+def int_field(
```
pbadillatorrealba, line 9: Maybe here you could try beartype to make sure that the types that are being sent to this function are the ones that should be being sent (runtime type checking).

```suggestion
# from beartype import beartype
# ...
@beartype
def int_field(
```

DashAI/back/core/schema_fields/int_field.py
```diff
@@ -0,0 +1,46 @@
...
+    Parameters
+    ----------
+    description: str
+        Description of the field.
+    default: int
+        The default value to show to the user.
+    minimum: Optional int
+        Minimum value of the field.
+    maximum: Optional int
+        Maximum value of the field.
```
pbadillatorrealba: By copilot:
```suggestion
    Parameters
    ----------
    description: str
        A string that describes the field.
    default: int
        The default integer value that will be displayed to the user.
    ge: Optional[int]
        An optional integer that the value should be greater than or equal to.
        If not provided, there is no lower limit.
    gt: Optional[int]
        An optional integer that the value should be strictly greater than.
        If not provided, there is no strict lower limit.
    le: Optional[int]
        An optional integer that the value should be less than or equal to.
        If not provided, there is no upper limit.
    lt: Optional[int]
        An optional integer that the value should be strictly less than.
        If not provided, there is no strict upper limit.
```

DashAI/back/core/schema_fields/int_field.py
```diff
@@ -0,0 +1,46 @@
...
+    """
+    params = {"description": description, "default": default}
+    if minimum:
+        params["ge"] = minimum
+    if maximum:
+        params["le"] = maximum
+    return Annotated[
+        int,
+        Field(**params),
+    ]
```
pbadillatorrealba: Here mypy is claiming to me that the types do not match. How about passing it the arguments directly (which by type checking, should always match or be None (which as I saw in the field docstring, is a possible value for the arguments))?

```suggestion
    return Annotated[
        int,
        Field(
            description=description,
            default=default,
            ge=ge,
            gt=gt,
            le=le,
            lt=lt,
        ),
    ]
```

DashAI/back/core/schema_fields/component_field.py
```diff
@@ -0,0 +1,46 @@
+from pydantic import BaseModel, Field
+from typing_extensions import Annotated
+
+
+class ComponentType(BaseModel):
+    component: str
+    params: dict
+
+
+def component_field(
```
pbadillatorrealba, line 12: Same here, you could use beartype to check if the parameter matches the specified types.

```suggestion
# from beartype import beartype
#....
@beartype
def component_field(
```

DashAI/back/core/schema_fields/string_field.py
```diff
@@ -0,0 +1,65 @@
...
+    """
+
+    def check_str_in_enum(x: str) -> str:
+        assert x in enum, f"{x} is not in the enum"
+        return x
+
+    return check_str_in_enum
+
+
+def string_field(
```
pbadillatorrealba, line 33:
```suggestion
# from beartype import beartype
#....
@beartype
def string_field(
```

DashAI/back/core/schema_fields/utils.py
```diff
@@ -0,0 +1,44 @@
...
+
+    This function transforms all fields of the component into actual components.
+    To do this, the component type is looked up in the component registry and
+    instantiated using the corresponding parameters.
+
+    The function is called recursively on component parameters, to fill its
+    internal components.
+
+    Parameters
+    ----------
```
pbadillatorrealba: This function looks a bit complex. 
Could you write an example in the docstring (similar to the one in the component registry?)

```suggestion
    The function is called recursively on component parameters, to fill its
    internal components.
    
    Example
    ---------
    ...

    Parameters
    ----------
```

pbadillatorrealba (changes requested):
Second batch of comments :)

DashAI/back/core/schema_fields/string_field.py
Riul1999: Yes, unless we set the validate_default param to True in the Field function. I will add it to string and int fields, the component field it's quite different.

DashAI/back/core/schema_fields/string_field.py
Riul1999: Ok, i used assert beacause is used in the pydantic examples for validators, but if it can be removed is better to use Exceptions.

DashAI/back/core/schema_fields/component_field.py
Riul1999: Moved to the notion task description.

DashAI/back/core/schema_fields/utils.py
Riul1999: Yes it's working, test_schemas::test_normal_schema checks that in the init of the components, the inits checks that the objects are a instance of the class and not the raw dictionary.

DashAI/back/core/schema_fields/int_field.py
Riul1999: Ok, I used minimum and maximum to follow the names used in the schematic, but it is more explicit this way.

DashAI/back/core/schema_fields/int_field.py
Riul1999: It's a better option, i use the dictionary because i wasn't sure that the Field function could handle the Nones.

DashAI/back/core/schema_fields/component_field.py
```diff
@@ -0,0 +1,41 @@
...
+    Returns
+    -------
+    type[ComponentType]
+        A pydantic-like type to represent the component.
+    """
+    return Annotated[
+        ComponentType,
+        Field(
+            description=description,
+            default=default,
```
cristian-tamblay: Are we leaving the default option then? We could also merge this to the project branch and then fix the "default problem". I think something like having `placeholder=placeholder` would work better than a default, because that would not remove it from the `required` field

Riul1999: No, the default will be changed into placeholder, but in another PR. There is a Task for that in Notion https://www.notion.so/dash-ai/Change-default-to-placeholder-a22d7e73a10645c8a0a25205ad845a29?pvs=4

Riul1999: Yes., that's the idea. I create [this](https://www.notion.so/dash-ai/Change-default-to-placeholder-a22d7e73a10645c8a0a25205ad845a29?pvs=4) Notion Task to solve it later.

pbadillatorrealba (approved):

PR #181, opened 2024-02-01
NWFT Meta-Model

Riul1999:
# Summary

BagOfWordsTextClassificationModel has been resurrected with the new pydantic-like Json Schemas 😮.

## Type of change

- Back end new feature.

## Changes

- Move and Fix BagOfWordsTextClassificationModel model, now it uses a DashAIDataset as input and the tabular_classfier given in the model params. It vectorize all the training samples and uses a TabularClassificationModel to classify the input vectorized text.
- Add BagOfWordsTextClassificationModel model into the component registry.
- Add a minimal test to check the model is working well. Please remeber to delete this test before merging the changes, so as not to slow down the execution of the tests.

## How to Test

You can test this changes using ``pytest tests/back/models/test_bow_text_class_model.py``.
You can check that the model is in the registry by requesting the enpoint ``/component/&id=BagOfWordsTextClassificationModel``.

## Notes

Please do not forget to delete the test before merging this PR.
This PR is blocked by https://github.com/DashAISoftware/DashAI/pull/179

PR #182, opened 2024-02-01
Pydantic Schemas for all Components

Riul1999:
# Summary

Create pydantic Schemas for all components and delete deprecated json schema files.

## Type of change

- Refactoring.

## Changes

- Delete JSON schema and create pydantic Schema for:
  - **DecisionTreeClassifier**
  - **DistilBertTransformer**
  - **DummyClassifier**
  - **HistGradientBoostingClassifier**
  - **KNeighborsClassifier**
  - **LogisticRegression**
  - **OpusMtEnESTransformerSchema**
  - **RandomForestClassifier**
  - **SVC**
  - **ViTTransformer**
- Add new field types: **float** and **bool**

## How to Test

You can check that the component JSON schema by requesting the enpoint ``/component/&id={component_name}``.

## Notes

This PR is blocked by https://github.com/DashAISoftware/DashAI/pull/181


DashAI/back/core/schema_fields/float_field.py
```diff
@@ -0,0 +1,53 @@
...
+from pydantic import Field
+from typing_extensions import Annotated
+
+
+def float_field(
+    description: str,
+    default: float,
+    minimum: Optional[float] = None,
+    exclusive_minimum: Optional[float] = None,
+    maximum: Optional[float] = None,
```
pbadillatorrealba:
```suggestion
    ge: Optional[float] = None,
    gt: Optional[float] = None,
    le: Optional[float] = None,
    lt: Optional[float] = None,
```

DashAI/back/core/schema_fields/float_field.py
```diff
@@ -0,0 +1,53 @@
...
+    description: str
+        Description of the field.
+    default: float
+        The default value to show to the user.
+    minimum: Optional float
+        Minimum value of the field.
+    exclusive_minimum: Optional float
+        Exclusive minimum value of the field.
+    maximum: Optional float
+        Maximum value of the field.
```
pbadillatorrealba:
```suggestion
    ge: Optional[float]
        An optional float that the value should be greater than or equal to.
        If not provided, there is no lower limit.
    gt: Optional[float]
        An optional float that the value should be strictly greater than.
        If not provided, there is no strict lower limit.
    le: Optional[float]
        An optional float that the value should be less than or equal to.
        If not provided, there is no upper limit.
    lt: Optional[float]
        An optional float that the value should be strictly less than.
        If not provided, there is no strict upper limit.
```

DashAI/back/core/schema_fields/float_field.py
```diff
@@ -0,0 +1,53 @@
...
+    if minimum:
+        params["ge"] = minimum
+    if exclusive_minimum:
+        params["gt"] = exclusive_minimum
+    if maximum:
+        params["le"] = maximum
+    return Annotated[
+        float,
+        Field(**params),
+    ]
```
pbadillatorrealba:
```suggestion
    return Annotated[
        int,
        Field(
            description=description,
            default=default,
            ge=ge,
            gt=gt,
            le=le,
            lt=lt,
        ),
    ]
```

pbadillatorrealba (changes requested):
The changes are looking great (really)!

I left the same comment as in the previous pr. 
If you change to ge, gt, lt, le for int, then the float field should also reflect that change.

Regarding the rest of the changes, elimination of the json and schemas inside the models, I think it's great!

pbadillatorrealba (approved):

PR #183, opened 2024-02-07
Plugins API

Riul1999:
# Summary

Add endpoints to manage a plugin.

## Type of change

- Back end new feature.

## Changes

- Add _plugin_ route to api_v1.
- The implemented methods are:
  - GET _/_ : endpoint to get all available plugins in the DB. Add plugin_status and tags filters.
  - GET _/{plugin_id}_ : enpoint to get an especific plugin.
  - POST _/_ : endpoint to load a list of raw_plugins into the local DB. If the plugin is registered in the DB the record is updated.
  - POST _/index_: endpoint to get all the available plugin in PyPI and load them into the local DB. If the plugin is registered in the DB the record is updated.
  - DELETE _/{plugin_id}_ : endpoint to delete a plugin, the tags related to the plugin are also deleted.
  - PATCH _/{plugin_id}_ : endpoint to update the status of the plugin.
- Add PluginParams, TagParams and PluginUpdateParams for validate the input of the api.
- Add PluginStatus enum to register the state of the plugin, can be REGISTERED, DOWNLOADED, INSTALED or ERROR.
- Add PluginTag enum to register the tag assoc to a plugin, can be DashAI, Package, Task, Model, Dataloader, Converter or Explainer. 
- Create Plugin and Tag tables in the DB, to persist the entities.
- Add mock function in _plugin/utils.py_ to retrieve the data of the plugins from PyPI.
- Add test to check the correct behaving of the endpoints.

## How to Test

Run ``pytest \tests\back\api\test_plugins.py``


DashAI/back/api/api_v1/endpoints/plugins.py
```diff
@@ -0,0 +1,284 @@
...
+    with session_factory() as db:
+        logging.debug("Storing plugin metadata in database.")
+        raw_tags = raw_plugin.tags
+        try:
+            plugin = Plugin(
+                name=raw_plugin.name,
+                author=raw_plugin.author,
+                summary=raw_plugin.summary,
+                description=raw_plugin.description,
+                description_content_type=raw_plugin.description_content_type,
```
pbadillatorrealba: Does this field meet any kind of specification? Something like html content type?

DashAI/back/api/api_v1/endpoints/plugins.py
```diff
@@ -0,0 +1,284 @@
...
+            raise HTTPException(
+                status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+                detail="Internal database error",
+            ) from e
+
+    return plugin
+
+
+@inject
+def add_plugin_to_db(
```
pbadillatorrealba: This function is not associated with any API. Is this OK?

DashAI/back/api/api_v1/endpoints/plugins.py
```diff
@@ -0,0 +1,284 @@
...
+    params : List[PluginParams]
+        The new plugins parameters.
+
+    Returns
+    -------
+    List[Plugin]
+        A list with the created plugins.
+    """
+    with session_factory() as db:
+        logging.debug("Storing plugin metadata in database.")
```
pbadillatorrealba: Use local logger, not the global one (logging)

```suggestion
        logger.debug("Storing plugin metadata in database.")
```

DashAI/back/core/enums/status.py
```diff
@@ -7,3 +7,10 @@ class RunStatus(Enum):
     STARTED = 2
     FINISHED = 3
     ERROR = 4
+
+
+class PluginStatus(Enum):
+    REGISTERED = 0
+    DOWNLOADED = 1
+    INSTALED = 2
+    ERROR = 3
```
pbadillatorrealba: Maybe we could leave error as 99 (to avoid moving it)?

DashAI/back/api/api_v1/endpoints/plugins.py
```diff
@@ -0,0 +1,284 @@
...
+            )
+            db.add(plugin)
+            db.commit()
+            db.refresh(plugin)
+
+            for raw_tag in raw_tags:
+                tag = Tag(
+                    plugin_id=plugin.id,
+                    name=raw_tag.name,
+                )
```
pbadillatorrealba: What if this relationship already exists?

DashAI/back/api/api_v1/endpoints/plugins.py
```diff
@@ -0,0 +1,284 @@
...
+    Returns
+    -------
+    List[Plugin]
+        A list with the created plugins.
+    """
+    plugins = [add_plugin_to_db(param) for param in params]
+    return plugins
+
+
+@router.post("/refresh", status_code=status.HTTP_201_CREATED)
```
pbadillatorrealba: This API is incorrectly designed.  The idea is that nouns, not verbs, live within the API definition. Meanwhile, the actions (verbs) should always be get, post, put, delete, patch, etc...

Here you need to make a call to a subresource. Maybe it would be a good idea to make a get call to `plugin/index` or `plugin/db`?

DashAI/back/dependencies/database/models.py
```diff
@@ -99,3 +100,39 @@ def set_status_as_finished(self) -> None:
...
+
+class Tag(Base):
+    __tablename__ = "tag"
+    """
+    Table to store all the tags related to a plugin
+    """
+    id: Mapped[int] = mapped_column(primary_key=True)
+    plugin: Mapped["Plugin"] = relationship(back_populates="tags")
+    plugin_id: Mapped[int] = mapped_column(ForeignKey("plugin.id"))
+    name: Mapped[Enum] = mapped_column(Enum(PluginTag), nullable=False)
```
pbadillatorrealba, line 138: This complicates me a little. What will happen if we dynamically add more plugins, for example, when a new model type is installed that is taged with something that did not exist in the original enum? 
I think we should not have an enum at all, but just leave this field open.

So we could also have tags like Scikit, HuggingFace, SHAPLike or others.

DashAI/back/plugins/utils.py
```diff
@@ -0,0 +1,44 @@
...
+
+    Returns
+    ----------
+    List[str]
+        A list with the names of all PyPI packages
+    """
+    return [
+        "dashai-tabular-classification-package",
+        "pytorch",
+        "sklearn",
```
pbadillatorrealba, line 16: What is this?
If these are the possible names of the plugins, I don't think it's a good idea that they don't have the dashai prefix.

DashAI/back/plugins/utils.py
pbadillatorrealba, line 1: What is the purpose of this module and its functions?

tests/back/api/test_plugins.py
```diff
@@ -0,0 +1,85 @@
...
+        ],
+    )
+    assert response.status_code == 201, response.text
+    assert len(response.json()) == 1
+
+
+def test_refresh_plugins(client: TestClient):
+    response = client.post("/api/v1/plugin/refresh")
+    assert response.status_code == 201, response.text
+    assert len(response.json()) == 1
```
pbadillatorrealba: Maybe here in the future we could test the refresh using pypi test as index.

DashAI/back/api/api_v1/endpoints/plugins.py
```diff
@@ -0,0 +1,284 @@
...
+                status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+                detail="Internal database error",
+            ) from e
+
+    return plugins
+
+
+@router.get("/{plugin_id}")
+@inject
+async def get_plugin(
```
pbadillatorrealba, line 75: Since we have tags, I think it would be prudent to also have the option to bring some plugins according to their tag or a set of tags ( by making a union of sets).

DashAI/back/api/api_v1/endpoints/plugins.py
pbadillatorrealba: Another option would be to have another API, something like plugins-db where this type of requests are made.

DashAI/back/plugins/utils.py
pbadillatorrealba: If this is decided, there might even be somewhere some check that the plugins name starts with dashai.

pbadillatorrealba (commented):
Very good progress!

I left some comments regarding the API and some implementation doubts.

DashAI/back/api/api_v1/endpoints/plugins.py
pbadillatorrealba: Also, wouldn't it be nice to be able to filter by status?
This could make it much easier to get the plugins that are installed in a view, eventually designed to manage them.

DashAI/back/plugins/utils.py
pbadillatorrealba: I read the following pr, then I understood what this module was about. Closing comment.

DashAI/back/api/api_v1/endpoints/plugins.py
Riul1999: I have added an if to check if the tag already exists, if the tag exists the program keeps trying to add another tag from the tag list.

DashAI/back/plugins/utils.py
Riul1999: This is to show that the function filters out plugins that do not start with the prefix "dashai-".

DashAI/back/api/api_v1/endpoints/plugins.py
Riul1999: No, i will move it to dependencies/database and import it. Thanks.

DashAI/back/api/api_v1/endpoints/plugins.py
Riul1999: Yes, the field format is the same as the HTTP ContentType header. For now the only content-type supported is "text/markdown", maybe in the future we will enhance it to support other content-types.

DashAI/back/dependencies/database/models.py
Riul1999: At the beginning I proposed to use an enum to avoid the creation of several same tags but with typographical errors, in order to force the plugin developer to check that the declared plugins are well typed. 
But now I think this is too restrictive, at this stage I think it is better to be more flexible, I will create a task in notion to think about it in the future.

DashAI/back/api/api_v1/endpoints/plugins.py
Riul1999: I think the first option is better, we still don't have enough functionality to make two endpoints.
I will rename it to POST plugin/index, to the endpoint to upload all the plugins that exist in the index (PyPI).

pbadillatorrealba (approved):
LGTM!

PR #184, opened 2024-02-08
Change default for placeholder

Riul1999:
# Summary

To prevent the JSON schema for omitting mandatory fields, it's necessary to avoid the use of default keyword in fields, so it's proposed to use placeholder keyword.

## Type of change

- Bug fix.

## Changes

- Use placeholder instead of default in field type functions. The placeholder keyword it's added to the JSON schema using the json_schema_extra param of Field function.
- Refactor all Schemas to use placeholder instead of default.
- Add test to check that the placeholder does not hide the mandatory fields.

## How to Test

Run `pytest .\tests\back\schema\test_schemas.py` and `pytest .\tests\back\models\test_numerical_wrapper.py`


pbadillatorrealba:
What is the objective of this change?

I am a little concerned because in the front end, placeholder is known as the text that suggests a text in an input, but does not represent a value. 

Placeholder:
image


If we use placeholder instead of default, it is likely that the user will have to fill in all the parameters by hand. 

I think that's not a good idea :(

Riul1999:
> What is the objective of this change?
> 
> I am a little concerned because in the front end, placeholder is known as the text that suggests a text in an input, but does not represent a value.
> 
> Placeholder: image
> 
> If we use placeholder instead of default, it is likely that the user will have to fill in all the parameters by hand.
> 
> I think that's not a good idea :(

We made this change because if we pass a default value to schemas, pydantic assumes that the field is not required. So we renamed it to placeholder to avoid this problem. 

We chose the name placeholder because it is the value that is shown to the user when they go to fill in the fields. We can rename it but I don't know a more appropriate name.

pbadillatorrealba:
> We made this change because if we pass a default value to schemas, pydantic assumes that the field is not required. So we renamed it to placeholder to avoid this problem.

But, this is a problem only when the default value is None? 

If so, what is the objective of having a None default value?

Riul1999:
> > We made this change because if we pass a default value to schemas, pydantic assumes that the field is not required. So we renamed it to placeholder to avoid this problem.
> 
> But, this is a problem only when the default value is None?
> 
> If so, what is the objective of having a None default value?

No, the "problem" is that the generated schema has a field called _required_ that specifies the fields that the user must fill in to generate the model (the schema) and if we pass a default to the Field function, pydantic thinks that the field is no longer required and will not throw an error if we don't pass the field as parameter, which is a bad behavior since we expect the validation to fail if we skip a field.
To avoid this we rename the field to placeholder, as I said before maybe the name could be changed to be clearer.

pbadillatorrealba (approved):
Ok, let's see how it works :)
When we make the documentation, we must make explicit that placeholder is used to validate default values.

PR #185, opened 2024-02-13
Custom Union and Optional fields

Riul1999:
# Summary

Added union and null field types to create schema union and nullable fields.

## Type of change

- Back end new feature.
- Refactoring.

## Changes

- Create _schema_field_ funtion to create a schema field, it receives the type of the field, the **placeholder** and the **description**.
- Create union and optional types that simple add a **Union** o **Optional** type in front of the actual type(s).
- Refactor model schemas to use _schema_field_.
- Update test_schemas to use _schema_field_ and test the generated json schema of the **UnionSchema**.
- Refactor _string_field_ enum param, now it's added by modifying the json schema directly instead of using the _json_schema_extra_ param of Field function.

## How to Test

Run ``pytest \tests\back\schemas\test_schemas.py``


DashAI/back/core/schema_fields/string_field.py
```diff
@@ -28,19 +30,42 @@ def check_str_in_enum(x: str) -> str:
...
+        ) -> JsonSchemaValue:
+            json_schema = handler(core_schema)
+            json_schema = handler.resolve_ref_schema(json_schema)
+            json_schema["enum"] = enum
+            return json_schema
+
+    return StringFieldWithEnum
+
+
+def string_field(enum: List[str]) -> Type[str]:
```
pbadillatorrealba, line 64: Question: For the case of a component that requires names, for example, a dataset, this solution would not allow you to add arbitrary names since it would only allow strings. 
Is there a possibility of this happening?
And if it happens, would it be necessary to implement a new type of string field?

pbadillatorrealba (commented):
schema_field will not be too much over-engineering?
Assuming that we implement all the possible fields ourselves, I don't think it is necessary to have a wrapper to put the placeholder and description to each of these.

Besides, I think it was much clearer when the parameterized Field was used directly.

The other thing is that I think it looks a little strange if you replace the fields:

For example, an int field is:

```python
t = Annotated[int, Field(ge=ge, gt=gt, le=le, lt=lt)].
```
Then, when passing through the schema_field, it would look like this: 

```python
Annotated[
        t,
        Field(description=description, json_schema_extra={"placeholder": placeholder}),
    ]
```

Which in the execution time would translate to 

```python
Annotated[
        Annotated[int, Field(ge=ge, gt=gt, le=le, lt=lt)],
        Field(description=description, json_schema_extra={"placeholder": placeholder}),
    ]
```


Which I think is not compatible with pydantic.

What do you think?

DashAI/back/core/schema_fields/string_field.py
Riul1999: Yes, this field only accepts a specific input, i think it's less cleaner to have a field that do the both things.
In PR #193 I rename this field to enum_field and create the string_field that accepts any string, specifically to accept the name of a dataset.

Riul1999:
> schema_field will not be too much over-engineering? Assuming that we implement all the possible fields ourselves, I don't think it is necessary to have a wrapper to put the placeholder and description to each of these.

I think the refactor is necessary, imagine we have a field that accepts a string or an int (a union field) if we use the old style int_field and string_field, we will have two placeholders and two descriptions, which placeholder should the front end choose? the description will have to be repeated in both fields? which field is more important? I think these changes give a better "shape" to the schema fields. Also with the optional field I think it is incorrect to pass a None as placeholder of an int_field.

> 
> Besides, I think it was much clearer when the parameterized Field was used directly.
> 
> The other thing is that I think it looks a little strange if you replace the fields:
> 
> For example, an int field is:
> 
> ```python
> t = Annotated[int, Field(ge=ge, gt=gt, le=le, lt=lt)].
> ```
> 
> Then, when passing through the schema_field, it would look like this:
> 
> ```python
> Annotated[
>         t,
>         Field(description=description, json_schema_extra={"placeholder": placeholder}),
>     ]
> ```
> 
> Which in the execution time would translate to
> 
> ```python
> Annotated[
>         Annotated[int, Field(ge=ge, gt=gt, le=le, lt=lt)],
>         Field(description=description, json_schema_extra={"placeholder": placeholder}),
>     ]
> ```
> 
> Which I think is not compatible with pydantic.
> 
> What do you think?

This is weird I know but pydantic seems to work well with this, it generates all the current schemes. So I think it is not a problem thinking about what we are gaining.

pbadillatorrealba (approved):
Ok, if it works with pydantic, then I don't see a problem!

Also, your explanation made it much clearer to me the justification for why it is necessary. Besides, it is also in the same line with the ideas of the component spec, which should not be too different from this :)

Approved.

PR #186, opened 2024-02-13
fix: :bug: Refactor main function and add app_init module because of …

Lucas-CE:
# Summary

When training a model, the backend throw this error:

```
RuntimeError: Task <Task pending name='Task-64' coro=<RequestResponseCycle.run_asgi() running at C:\Users\lucas\python-envs\dashai-dev\lib\site-packages\uvicorn\protocols\http\httptools_impl.py:426> cb=[set.discard()]> got Future <Future pending> attached to a different loop
```

The changes solve this problem, but when running the backend, it doesnt wait for the timer to open the browser.

### DIFF
--- a/DashAI/__main__.py
+++ b/DashAI/__main__.py
@@ -1,12 +1,8 @@
-import pathlib
 import threading
 import webbrowser
 
 import typer
 import uvicorn
-from typing_extensions import Annotated
-
-from DashAI.back.app import create_app
 
 
 def open_browser():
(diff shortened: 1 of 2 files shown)


pbadillatorrealba:
Didn't you try the other option, to leave the loop parameter set to none?

I don't really like the idea of leaving the app in a separate file. If for some reason, this file is imported during the execution of a test, the server will be started in vain.
Also, I don't know what side effects declaring the app in this file at runtime might have.

If you prove that this file is not executed during tests, I approve this pr. But I would like you to see the second option first.

Edit: Also, there is some test that didn't finished. Check what happened to them.

PR #187, opened 2024-02-14
Feat/pypi data retrieval

Lucas-CE:
# Summary

Functions and tests implementation to pypi data retrieval.

- Plugins list retrieved from pypi aren't in lower case.
- Plugin names could start with DashAI or dashai -> get_plugins_from_pypi check if the name.lower() starts with dashai
- JSON Request doesn't care if you use lowercase or uppercase.


DashAI/back/plugins/utils.py
```diff
@@ -10,35 +14,30 @@ def _get_all_plugins() -> List[str]:
...
-    ]
+
+    client = xmlrpc.client.ServerProxy("https://pypi.python.org/pypi")
+    # get a list of package names
+    packages = client.list_packages()
+
+    return packages
 
 
 def _get_plugin_data(plugin_name: str) -> dict:
```
pbadillatorrealba: Let's try to use the same names we have been using so far!

```suggestion
def _get_plugin_by_name_from_pypi(plugin_name: str) -> dict:
    # Also add a docstring!
```

DashAI/back/plugins/utils.py
```diff
@@ -10,35 +14,30 @@ def _get_all_plugins() -> List[str]:
...
     return raw_plugin
 
 
 def get_plugins_from_pypi() -> List[dict]:
-    plugins_names = filter(lambda name: "dashai" in name, _get_all_plugins())
+    plugins_names = [
+        plugin_name.lower()
+        for plugin_name in _get_all_plugins()
+        if plugin_name.lower().startswith("dashai") and plugin_name.lower() != "dashai"
+    ]
```
pbadillatorrealba, line 69: How long does this operation take?

DashAI/back/plugins/utils.py
```diff
@@ -10,35 +14,30 @@ def _get_all_plugins() -> List[str]:
...
+    )
+
+    raw_plugin: json = response.json()["info"]
+    raw_plugin["tags"] = [
+        {"name": keyword} for keyword in raw_plugin.pop("keywords").split(",")
+    ]
     return raw_plugin
 
 
 def get_plugins_from_pypi() -> List[dict]:
```
pbadillatorrealba, line 56: Document this function.

pbadillatorrealba (commented):
Reviewed.
I left minor comments ;)

pbadillatorrealba:
As in the previous pr, I am still in favor of restricting plugin names to only those that start with dashai. Otherwise, we are going to have a huge mess later.

DashAI/back/plugins/utils.py
Lucas-CE: The _get_all_plugins execution 4.84 secs.
The other operations 0.7 secs.

pbadillatorrealba:
Fix the conflicts and we are ready to merge!

PR #188, opened 2024-02-18
New Component Type Schema

Riul1999:
# Summary

Implemented new way to generate json schema for ComponentType fields.

## Type of change

- Refactoring.

## Changes

- Added function __component_type_factory_ to create a _ComponentType_ class parameterized by the parent field.
- Overrided __get_pydantic_core_schema__ to avoid the generation of $defs field in the json schema.
- Overrided __get_pydantic_json_schema__ to add parent field into json schema.
- Fix **NumericalWrapperForText** schema, update placeholder and fix typo in description.
- Update tests, check parent field and the new placeholders for component fields.

## How to Test

Run `pytest .\tests\back\schemas\test_schamas.py` and `pytest .\tests\back\models\test_numerical_wrapper.py`


DashAI/back/models/scikit_learn/bow_text_classification_model.py
```diff
@@ -22,7 +22,7 @@ class BagOfWordsTextClassificationModelSchema(BaseSchema):
 
     tabular_classifier: schema_field(
         component_field(parent="TabularClassificationModel"),
-        placeholder=100,
+        placeholder={"component": "SVC", "params": {}},
```
pbadillatorrealba, line 25: At this moment, because of the dependency injection issue, it is not verifying if the component really exists or not, right?
What would happen if the component is not found? What if it has no schema?

DashAI/back/models/scikit_learn/bow_text_classification_model.py
```diff
@@ -22,7 +22,7 @@ class BagOfWordsTextClassificationModelSchema(BaseSchema):
 
     tabular_classifier: schema_field(
         component_field(parent="TabularClassificationModel"),
```
pbadillatorrealba, line 24: In the future, do you think it will be possible to receive classes directly instead of their names?

pbadillatorrealba (approved):
The change looks great!
With this I guess the whole configurable object idea would be working well.

I left you a couple of queries in the code, but the pr is approved!

cristian-tamblay (approved):

PR #190, opened 2024-02-19
Validation flow refactoring

Riul1999:
# Summary

Modified the validation and transformation flow of the component parameters. Now these steps are executed by the component within it's init method.

## Type of change

- Refactoring.

## Changes

- Adds the _validate_and_transform_ method inside ConfigObject, to perform the validation and transformation of the component's params.
- Refactor the _fill_objects_ function so that it is not recursive, now it is not necessary.
- Refactor JobModel, now it is not in charge of validating and transforming the params of the model.
- Refactor all init methods in models.
- Refactor tests that used configurable components, now it's required to fill all the params.
- Refactor _test_schemas_ to make validation within dummy components.

## How to Test

Run `pytest tests\back\schemas\test_schemas.py` and `pytest tests\back\models\test_numerical_wrapper.py`


DashAI/back/config_object.py
```diff
@@ -16,3 +17,7 @@ def get_schema(cls) -> dict:
             Dictionary representing the Json Schema of the component.
         """
         return cls.SCHEMA.model_json_schema()
+
+    def validate_and_transform(self, raw_schema: dict) -> dict:
```
pbadillatorrealba: Pls document this function!

pbadillatorrealba: It also adds a note of the exception or error that should be raised when a data set is incorrectly validated against your schema.

pbadillatorrealba: Why not call the data `raw_data` instead of raw_schema?
```suggestion
    def validate_and_transform(self, raw_data: dict) -> dict:
```

pbadillatorrealba (changes requested):
I'm okay with the change!

I requested minor changes.

PR #192, opened 2024-02-22
Plugin front views and API

fvalladaresj:
# Summary

Implement front end for plugins system.

## Type of change

- Front end new feature.

## Changes

- Add plugins page
- Add plugins components
- Add pluginStatus enum constant to register the state of the plugin
- Add custom hooks for api connection
- Add sample data for json-server testing
- Add react-markdown library 

## How to Test

...

## Screenshots

...

## Notes

Includes changes from feat/plugins-api branch.


pbadillatorrealba:
@fvalladaresj , sorry for the delay, but I've already started reviewing your pr. 

Could you tell me how to execute your changes?
In particular, if you are using sample_db and how can I make it show up in the front end?

Also, tell me if I only have to check only the front code or check everything?

fvalladaresj:
@pbadillatorrealba I've updated my branch with the current project/plugins branch and fixed a mistake in the `getplugins` endpoint. Also, I'm not using `sample_db` anymore. 

To test, I follow these steps:
-  From DashAI/front, `yarn install` then `yarn start`.
- From DashAI, `pip install .` then `dashai` to run the backend.

Only the front files need reviewing since the back is now up to date with the project/plugins branch.

DashAI/front/src/types/plugin.ts
```diff
@@ -0,0 +1,19 @@
+import type { PluginStatus } from "../pages/plugins/constants/pluginStatus";
+
+interface Tag {
```
pbadillatorrealba: The naming convention was missing! It should be called ITag.

```suggestion
interface ITag {
```

DashAI/front/src/types/plugin.ts
```diff
@@ -0,0 +1,19 @@
...
+  id: number;
+  name: string;
+  plugin_id: number;
+}
+export interface IPlugin {
+  id: number;
+  name: string;
+  author: string;
+  tags: Tag[];
```
pbadillatorrealba: Since you already declared the tags as inteface, wouldn't it be a good idea that the tag array also be of type tag?
```suggestion
  tags: Array<ITag>];
```

requirements.txt
```diff
@@ -16,6 +16,6 @@ typer>=0.9.0
 rich>=13.5.3
 torch==1.13.0 #+cu116
 --find-links https://download.pytorch.org/whl/torch_stable.html
-transformers>=4.23.1
-sacrebleu>=2.3.1
-sentencepiece>=0.1.97
+transformers==4.23.1
+sacrebleu==2.3.1
+sentencepiece==0.1.97
```
pbadillatorrealba: What are the consequences of fixing those requirements?
I would prefer that, if it is not necessary to fix them, that we leave them as they were before.

However, it's not critical, as you will eventually outsource the requirements to each plugin. If you don't want to do it, no problem.

DashAI/front/src/pages/plugins/constants/pluginStatus.ts
pbadillatorrealba, line 1: Is it OK to have this file, on which the interface here depends? 
Wouldn't it be better to formalize the plugin states in the same typescript file?

pbadillatorrealba (changes requested):
Ok, sorry again for the delay.

I left you some minor comments regarding types and requirements.

I ran the front end, but I could see only the plugins administration interface, but none in particular, since I don't have any in the db.

Anyway, and according to what I saw in your code and in the multiple presentations you have made, I think the front end is in good shape to be merged!

Tell me about the comments I left you and then we can approve.

fvalladaresj:
@pbadillatorrealba  I merged PluginStatus enum into the Plugin type file and fixed all the imports. Also fixed Tag to ITag to follow conventions.

requirements.txt
fvalladaresj: This changes where made by @cristian-tamblay.

PR #202, opened 2024-04-30
Script to create python enviroment and run DashAI

Lucas-CE:
# Summary

This pull request introduces a script to automate the creation of a python virtual environment, installing dependencies, and running DashAI frontend and backend.

This automation only works if the PATH environment variable with the name 'python' points to a Python 3.8 executable. Given that when creating the environment the command python -m venv name is used in the script.


pbadillatorrealba (commented):
I think it is not a good practice to include this kind of scripts in the repository.
Instead, I think this file should be transformed into a section of the readme, or failing that, update the existing one with these instructions.
Now, if it is something you want to keep, I would recommend moving the script to a specific script folder, something like DashAI/scripts.

PR #211, opened 2024-05-28
Dependency Injection V2

pbadillatorrealba:
# Summary

This pr aims to change the current unmaintained python-dependency-injection to a new simpler one [kink](https://github.com/kodemore/kink).

To do so, this pr presents two main changes:
1. Change every python-dependency-injection occurence to the new kink based, using its simple lambda interface.
2. Change the test structure to use monkeypatch + kink instead of the override provided by the old di framework.

## Type of change

- Back end new feature.

## Changes

- Change every component registry dependency call from:

```python
    component_registry: ComponentRegistry = Depends(
        Provide[Container.component_registry]
    ),

```

to 

```python
    component_registry: ComponentRegistry = Depends(lambda: di["component_registry"]),
```

And the same for configs, database session factory and queue.

- Implement a new container based on kink framework instructions.
- Change DI overrides to container monkeypatch in every test suits that require the app running.

## How to Test

Run every avialable test

```
$ pytest .
```

Also, check the common flow of dataset loading, train and explainations.


cristian-tamblay (approved):
This is working as intended.

PR #218, opened 2024-06-11
Feat/uninstall plugin

inunezn7:
# Summary

This branch allows the user to uninstall one plugin. This includes removing it from the Python packages directory (by using pip uninstall) and the components registry. This implies modifying the patch plugin API to make it a bit more general because before these changes it just works to update the status param to "installed".

## Changes

- Plugin statuses are reduced to just two possible ones: Registered or Installed
- Tests related to the deleted status "Downloaded" were removed
- Uninstalling a plugin uninstalls it from the Python env (pip uninstall ...)
- Uninstalling a plugin removes it from the component registry: added methods to unregister component classes and also remove them from the relationship manager.

## How to Test

To install and uninstall a plugin it is suggested to see branch [feat/empty-container-for-plugins](https://github.com/DashAISoftware/DashAI/tree/feat/empty-container-for-plugins). This runs DashAI with no components registered from the backend on it. Then go to the "Plugins" tab, press the "refresh" button and install dashai-test2-tabular-classification-package


DashAI/back/dependencies/registry/component_registry.py
```diff
@@ -207,6 +211,35 @@ def register_component(self, new_component: Type) -> None:
...
+        base_type = self._get_base_type(component)
+
+        try:
+            self._registry[base_type].pop(component.__name__)
+            logger.info(f"Component removed: {component.__name__}")
+        except KeyError as e:
+            logger.error(
+                f"Error: Component named {component.__name__} does not exist "
+                f"in the registry. Exception: {e}"
+            )
```
pbadillatorrealba: Ideally, the same exception should be raised here to avoid suppressing the exception and making it go unnoticed.

```suggestion
        except KeyError as e:
            raise ValueError(
                f"Error when unregistering a component: Component named {component.__name__} does not exist "
                f"in the registry. Exception: {e}"
            )

```

DashAI/back/dependencies/registry/relationship_manager.py
```diff
@@ -66,6 +70,43 @@ def add_relationship(
...
+            )
+
+        try:
+            self._relations[second_component_id].remove(first_component_id)
+        except KeyError as e:
+            logger.error(
+                f"Error: Relationship between {second_component_id} and does "
+                f"not exist {first_component_id} in the registry. Exception: "
+                f"{e}"
+            )
```
pbadillatorrealba: Same as before: here you should use ValueError to report the error. 
Just to comment: these operations in a normal operation should always work, so if this case occurs, it implies that something is wrong and therefore, it should be notified by means of an exception.

DashAI/back/plugins/utils.py
```diff
@@ -102,38 +102,7 @@ def get_available_plugins() -> List[type]:
...
-            component_registry.register_component(plugin)
-        except Exception as e:
-            logging.exception(e)
-    return list(new_plugins)
-
-
-def install_plugin_from_pypi(pypi_plugin_name: str) -> None:
+def install_plugin_from_pypi(pypi_plugin_name: str, install: bool = True) -> None:
     """
     Register only new plugins in component registry
```
pbadillatorrealba: It wouldn't be better to call this method `call_pip` or `execute_pip_command` and set the install param as operation, which could take `install` or `uninstall` strings as values?
Also, remember to update the method docstring.

DashAI/back/plugins/utils.py
pbadillatorrealba, line 1: If I am not mistaken, this file contains all the logic of the installation and handling of the plugins. So, wouldn't it be a good idea to move it to the `core` folder?

DashAI/back/plugins/utils.py
```diff
@@ -147,12 +116,91 @@ def install_plugin_from_pypi(pypi_plugin_name: str) -> None:
...
+    installed_plugins = set(get_available_plugins()) - set(pre_installed_plugins)
+    return installed_plugins
+
+
+def register_plugin_components(installed_plugins, component_registry):
+    for plugin in installed_plugins:
+        try:
+            component_registry.register_component(plugin)
+        except Exception as e:
+            logging.exception(f"Plugin {plugin} could not be registered. Error: {e}")
```
pbadillatorrealba: The same about exceptions here. Here I would simply call the `register_component` method directly and have the method throw the exception in case it detects errors.

DashAI/back/plugins/utils.py
```diff
@@ -147,12 +116,91 @@ def install_plugin_from_pypi(pypi_plugin_name: str) -> None:
...
+    plugin_name : str
+        A string with the name of the plugin in pypi to install
+
+    component_registry : ComponentRegistry
+        The current app component registry
+
+    """
+    available_plugins: List[type] = get_available_plugins()
+    install_plugin_from_pypi(plugin_name, False)
+    uninstalled_components: List[type] = set(available_plugins) - set(
```
pbadillatorrealba, line 207: Here you are retrieving the set of the installed components!
```suggestion
    installed_components: List[type] = set(available_plugins) - set(
```

DashAI/back/plugins/utils.py
```diff
@@ -147,12 +116,91 @@ def install_plugin_from_pypi(pypi_plugin_name: str) -> None:
...
+        get_available_plugins()
+    )
+    return uninstalled_components
+
+
+def unregister_plugin_components(
+    uninstalled_plugins: List[type],
```
pbadillatorrealba: Those plugins are the plugins that will be removed. I think it is best to assume that, at this step, they are not removed yet.
```suggestion
    plugins: List[type],
```

DashAI/back/plugins/utils.py
```diff
@@ -147,12 +116,91 @@ def install_plugin_from_pypi(pypi_plugin_name: str) -> None:
...
+    Returns
+    ----------
+    List[type]
+        A list of plugins' classes wanted to be removed from the component registry
+    """
+    for plugin in uninstalled_plugins:
+        try:
+            component_registry.unregister_component(plugin)
+        except Exception as e:
+            logging.exception(e)
```
pbadillatorrealba: Try to be specific with the exceptions you are trying to capture here. Do RuntimeException, ValueError, TypeError? Or are there also others like OSError?
You can specify the possible exceptions using the following syntax:

```suggestion
        except (ValueError, TypeError, RuntimeError) as e:
            logging.exception(e)
```
Also, as a question: do you want that, even if the process of uninstalling a plugin fails, the rest are still attempted to be uninstalled?

I would suggest that this should not happen, as it can end up in strange side effects, but it's not so terrible either. I leave that judgement up to you.

pbadillatorrealba: Use a file specific logger, not the global one:

```suggestion
        except Exception as e:
            logger.exception(e)
```

And in the top of the file:
```python
logging.basicConfig(level=logging.INFO)
logger = logging.getLogger(__name__)
```

DashAI/back/api/api_v1/endpoints/plugins.py
```diff
@@ -253,14 +255,20 @@ async def update_plugin(
...
                 )
+            if params.new_status == PluginStatus.INSTALLED:
+                installed_components = install_plugin(plugin_name)
+                register_plugin_components(installed_components, component_registry)
+            elif (
+                plugin.status == PluginStatus.INSTALLED
+                and params.new_status == PluginStatus.REGISTERED
+            ):
+                uninstalled_components = uninstall_plugin(plugin_name)
+                unregister_plugin_components(uninstalled_components, component_registry)
```
pbadillatorrealba, line 259: I don't think it happens at the moment, but it could happen that there are plugins that are dependent on others?
What would happen in that case with the uninstallation of the plugin?

And the other question also related to uninstallation: When you uninstall the plugin, you only uninstall the plugin package, but not its dependencies right. I think that by default pip does that, but better to be clear about it

pbadillatorrealba (changes requested):
I am fine with the changes!
I left some comments about error handling and also some other conceptual things about uninstalling plugins.

Also, update the branch to the latest changes before the merge.
Note that it is possible that you have to change some component registry calls to the new framework!

DashAI/back/plugins/utils.py
Lucas-CE: Are you sure? I think that we are taking the set of installed plugins pre-uninstallation, and then we are removing the available plugins post uninstallation.

`
uninstalled_components: List[type] = set(available_plugins) - set(
        get_available_plugins()
    )
`

DashAI/back/dependencies/registry/component_registry.py
Lucas-CE: Totally agree! But I have a question. Why we should throw ValueError instead of KeyError

DashAI/back/plugins/utils.py
Lucas-CE: Changed! The method name is execute_pip_command now, and the install or uninstall action depends on a string.

DashAI/back/plugins/utils.py
Lucas-CE: Fixed! To be sure, we should change that to 
`
component_registry.register_component(plugin)
`
and that call should throw and catch the exception, right? (the change has been made, if you want to check how it looks like)

DashAI/back/plugins/utils.py
Lucas-CE: Fixed!

DashAI/back/plugins/utils.py
Lucas-CE: Due to the change in the code:
Old:
```
try:
     component_registry.register_component(plugin)
except Exception as e:
     logging.exception(f"Plugin {plugin} could not be registered. Error: {e}")
```
Now:
```
component_registry.register_component(plugin)
```
This comment is no longer relevant.

Lucas-CE: Due to the change in the code:
Old:
```
try:
     component_registry.register_component(plugin)
except Exception as e:
     logging.exception(f"Plugin {plugin} could not be registered. Error: {e}")
```
Now:
```
component_registry.register_component(plugin)
```
This comment is no longer relevant, because we are no longer using the logging.

DashAI/back/plugins/utils.py
Lucas-CE: Indeed. All the functions to handle plugins are in this file. We could consider to change the location after make all the fixes.

DashAI/back/api/api_v1/endpoints/plugins.py
Lucas-CE: Uninstalling a plugin doesn't uninstall the dependencies.

PR #219, opened 2024-06-12
Feat/empty container for plugins

inunezn7:
# Summary

The goal of this branch is to have the option of launching DashAI with a container with no component classes from the backend inside. This is useful for testing the correct plugin installation from PyPI in an empty DashAI environment and, at the same time, it allows the use of the software by simply loading the component classes defined in the backend.

## Type of change

- Back end new feature.

## Changes

- container_type argument introduced to app initialization
- get_initial_components method to populate the component's container

## How to Test

Run DashAI by using:

```python -m DashAI --container-type plugins```


requirements.txt
pbadillatorrealba, line 1: Thanks hahaha!

pbadillatorrealba (changes requested):
Same here, remember to update the branch to dev, there are many changes that will conflict in this branch.

PR #228, opened 2024-08-13
Feat/upgrade plugin

inunezn7:
# Summary

An upgrade button is added to plugins in case there is a newer version of the plugin

## Type of change

- Back end plugins new feature.

## Changes

...

## Notes

...


DashAI/back/api/api_v1/endpoints/plugins.py
```diff
@@ -267,3 +267,53 @@ async def update_plugin(
                 status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
                 detail="Internal database error",
             ) from e
+
+
+@router.patch("/{plugin_id}/upgrade")
```
inunezn7, line 276: We could merge this method with the other patch and make a general one to update plugin parameters. What do you think?

DashAI/back/api/api_v1/endpoints/plugins.py
```diff
@@ -267,3 +267,53 @@ async def update_plugin(
...
+                raise HTTPException(
+                    status_code=status.HTTP_404_NOT_FOUND,
+                    detail="Plugin not found",
+                )
+            if plugin.status == PluginStatus.INSTALLED:
+                uninstalled_components = uninstall_plugin(plugin_name)
+                unregister_plugin_components(uninstalled_components, component_registry)
+            installed_components = install_plugin(plugin_name)
+            register_plugin_components(installed_components, component_registry)
+            db.commit()
```
inunezn7: An update of the installed plugin version registered in the DB is needed here. It would be nice if the plugin installation returned the installed version.

DashAI/back/dependencies/database/utils.py
```diff
@@ -58,6 +59,8 @@ def add_plugin_to_db(
                 plugin = Plugin(
                     name=raw_plugin.name,
                     author=raw_plugin.author,
+                    version=raw_plugin.version,
```
inunezn7: We should call this variable installed_version for a better understanding

DashAI/back/api/api_v1/endpoints/plugins.py
Lucas-CE: I think it is a better practice to keep it separate. In particular considering the open closed principle.

DashAI/back/api/api_v1/endpoints/plugins.py
Lucas-CE: Ready! Now, after upgrading a plugin the database update all the plugin info with the new info of the plugin.

DashAI/back/dependencies/database/utils.py
Lucas-CE: Solved!

PR #231, opened 2024-10-05
Feat/regression

Lucas-CE:
# Summary

This pull request introduces new regression models and metrics.

### New Regression Models and Metrics:

* Added new regression models (`GradientBoostingR`, `LinearRegression`, `LinearSVR`, `MLPRegression`, `RandomForestRegression`, `RidgeRegression`) and metrics (`MAE`, `RMSE`).


cristian-tamblay:
Reverted some changes that made only regressions work. We need cesar change in the front to allow different metrics so regression uses RMSE instead of accuracy

PR #268, opened 2025-08-07
feat/pipelines

CarlaHayler:
Added support for configurable nodes using a `configSchema` object defined in `nodes.json`. Each node now includes a `configType` field that determines how the configuration is handled:

- If `configType` is "custom", a specific modal must be implemented.
- If it's "generic", the node must include a `configSchema` with the structure of the form, and the frontend will generate the configuration automatically.

This allows creating new nodes without the need to implement a specific modal for each one.


cristian-tamblay (approved):
Nice, at least this makes this a bit more general

PR #270, opened 2025-08-21
fix: refactor import statements in ConfigureExplorersStep and NodeSid…

Irozuku:
## Fix: Resolve circular dependency in frontend imports

This pull request resolves a **circular dependency issue** in the frontend components by updating import statements to use more explicit and direct paths.  

### Fixes and improvements
- Fixed a circular import between `ConfigureExplorersStep.jsx` and `ExplorersTable.jsx` by changing the import of `ExplorersTable` to a direct path (`../ExplorationsTable`).  
- Updated the import of `getNodeHelp` in `NodeSidebar.jsx` to reference `./nodeHelp` explicitly instead of relying on a relative index, removing ambiguity.  

#### Before
image

#### After
image


cristian-tamblay (changes requested):
Nice PR

cristian-tamblay (approved):

PR #303, opened 2025-09-23
Update db migration configuration

Creylay:
# Summary

This PR (re)introduce database migration using alembic (automatic migrations). The main objective is to be able to update the database without deleting the current one. Migrations scripts are generated using alembic at dev command and are applied automatically at app startup. 

Two main features of this PR is that migrations are applied automatically on app startup and adds CI configuration to check database consistency between models and migrations scripts (ensure there is no db drift).

## Type of change

- Back end new feature.

## Changes
- update alembic configuration to run db migrations
- reset migrations (delete old ones and create an initial one)
- runs migrations at app startup
- adds CI configuration to check consitency between models and migrations. This will ensure that all PR's that try to merge to develop ensures that migrations are in order (if they modify database.py then they create the correspondent migration file). 
- Update README.rst with alembic / migration instructions. 

## How to Test

No tests, but CI is running fine :) 

## Notes
Unfortunately, all devs will have to re-create their dbs by deleting it an creating it again using the alembic migration command (o just running the app).


cristian-tamblay (approved):

cristian-tamblay:
Lets hope this is a good idea for the future

PR #324, opened 2025-10-07
Add cache-control headers to front_api

cristian-tamblay:
# Summary

Improves how the frontend `index.html` file is served by adding cache-control headers.  
This ensures that users always receive the latest version of the React build after a new release, avoiding the need to manually refresh with **Shift + F5**.

## Type of change

- Back end new feature.

## Changes

- Updated `DashAI/back/api/front_api.py` to add `Cache-Control`, `Pragma`, and `Expires` headers when serving `index.html`.  
- Prevents outdated frontend builds from being loaded from browser cache.

## How to Test

1. Build the frontend with `yarn build`.  
2. Run the backend (`python -m DashAI` or the built `.exe`).  
3. Open the app in your browser (`http://localhost:8000/app/`).  
4. Deploy a new build and reload — the updated frontend should now load automatically without needing a hard refresh.

## Notes

This change only affects the `index.html` route.  
Static assets (JS, CSS, images) remain cached normally since their filenames are hash-based.


DashAI/back/api/front_api.py
```diff
@@ -18,7 +17,12 @@ async def read_index(
 ):
     front_build_path = config["FRONT_BUILD_PATH"]
     index_path = Path(f"{front_build_path}/index.html").absolute()
-    return FileResponse(index_path)
+    # Avoids caching of index.html (forces reload of front on every request)
+    response = FileResponse(index_path)
+    response.headers["Cache-Control"] = "no-cache, no-store, must-revalidate"
```
Irozuku: I think we shouldn't use `no-store` since it prevents any caching, even in memory, which can hurt performance unnecessarily. `must-revalidate `is redundant since `no-cache` already ensures the client revalidates the response each time ("If you want caches to always check for content updates while reusing stored content, no-cache is the directive to use. It does this by requiring caches to revalidate each request with the origin server.") [link](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Cache-Control)

DashAI/back/api/front_api.py
```diff
@@ -18,7 +17,12 @@ async def read_index(
...
+    response.headers["Pragma"] = "no-cache"
```
Irozuku: This is a deprecated feature and should be avoided. [link](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Pragma)

Irozuku: Oops. Line 23

DashAI/back/api/front_api.py
cristian-tamblay: no-cache should be enough, will check next front update if it works, should only affect the initial load of index.html when you open the app

Irozuku (approved):

PR #333, opened 2025-10-17
Replace dynamic imports with component registry lookup in converter job class

Irozuku:
This pull request refactors the way converter classes are instantiated in the `converter_job.py` file. Instead of dynamically importing converter modules using string manipulation and `import_module`, the code now retrieves converter classes directly from a `component_registry` provided by dependency injection. This simplifies the instantiation logic and reduces the risk of import errors.

Key changes:

**Refactoring converter instantiation:**

* Removed the use of `import_module` and dynamic module path construction for importing converters, replacing it with a lookup in the injected `component_registry` for the converter class.

### DIFF
--- a/DashAI/back/job/converter_job.py
+++ b/DashAI/back/job/converter_job.py
@@ -1,6 +1,5 @@
 import logging
 import re
-from importlib import import_module
 from pathlib import Path
 from typing import Dict, List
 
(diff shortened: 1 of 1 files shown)


cristian-tamblay:
Thank you this fixes a lot of import problems in the executables too. I wonder if this exact problem is somewhere else. I  think some devs/students didnt get how to use the dependency injection and class instanciation, this is a clear case of this.

cristian-tamblay (approved):

PR #350, opened 2025-10-23
Update tokenization from BOW model to remove all specified columns instead of just 'text'

Irozuku:
This pull request updates how columns are removed during tokenization in the `fit` and `predict` methods of the `bow_text_classification_model.py` file. Instead of removing only the `"text"` column, it now removes all columns present in the dataset, making the tokenization process more robust and less dependent on hardcoded column names.

**Tokenization improvements:**

* Updated both the `fit` and `predict` methods to remove all columns (`x.column_names`) during tokenization rather than just the `"text"` column, ensuring compatibility with datasets that may have different or multiple column names. [[1]](diffhunk://#diff-e365bc76750c3d480b2bf0dd787fb15ba89539aa6308958ce0d8e934a06fb487L206-R206) [[2]](diffhunk://#diff-e365bc76750c3d480b2bf0dd787fb15ba89539aa6308958ce0d8e934a06fb487L215-R215)

### DIFF
--- a/DashAI/back/models/scikit_learn/bow_text_classification_model.py
+++ b/DashAI/back/models/scikit_learn/bow_text_classification_model.py
@@ -203,7 +203,7 @@ class BagOfWordsTextClassificationModel(TextClassificationModel, SklearnLikeMode
         input_column = x.column_names[0]
         self.vectorizer.fit(x[input_column])
         tokenizer_func = self.get_vectorizer(input_column)
-        tokenized_dataset = x.map(tokenizer_func, remove_columns="text")
+        tokenized_dataset = x.map(tokenizer_func, remove_columns=x.column_names)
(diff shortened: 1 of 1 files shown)


cristian-tamblay:
There is a conceptual problem here. Maybe we need to rethink this.
What if the dataset had something like this:
[
  {"text": "hello world", "feature_1": 0.7, "feature_2": 0.9},
  ...
]
Then we will remove feature_1 and feature_2 with this change.

Worse than that, if the dataset is like
[
  {"feature_1": 0.7, "feature_2": 0.9, "text": "hello world"},
  ...
]

the vectorizer will fail miserably because we are assuming the text column will be the first. I think we really need to do the change of handling the dataset on the datasets page and do the preproc there before running experiments

cristian-tamblay (changes requested):
Conceptual problems

PR #357, opened 2025-10-27
Fix split experiment

Irozuku:
This pull request improves the dataset splitting workflow in both the backend and frontend, with a focus on better error handling, validation, and user feedback for dataset splits in experiments. The main changes are the introduction of more precise validation and clearer error messages for both random and manual splits, as well as a minor backend fix to dataset index handling.

**Frontend validation and user experience improvements:**

* Added separate error states and messages for random and manual dataset splits, providing clearer feedback to users about split requirements and validation errors.
* Improved validation logic for random splits to allow for small floating point errors and require all splits to be greater than zero and sum to one.
* Enhanced manual split validation to ensure each split has at least one row, with immediate feedback when editing indices.
* Updated UI components to display error states and helper text directly under the relevant inputs for both split types, improving clarity and usability.

**Backend fix:**

* Refactored the backend to explicitly assign train, test, and validation indexes from the split configuration.

### DIFF
--- a/DashAI/front/src/components/experiments/SplitDatasetRows.jsx
+++ b/DashAI/front/src/components/experiments/SplitDatasetRows.jsx
@@ -44,7 +44,9 @@ function SplitDatasetRows({
     testDatasetPercentage > 0;
 
   const checkSplit = (train, validation, test) => {
-    return train + validation + test === 1;
+    const sum = train + validation + test;
+    const tolerance = 0.0001; // Allow small floating point errors
+    return Math.abs(sum - 1) < tolerance;
   };
 
   // handle rows numbers change state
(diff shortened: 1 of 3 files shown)


DashAI/front/src/components/experiments/SplitDatasetRows.jsx
```diff
@@ -56,79 +58,137 @@ function SplitDatasetRows({
...
+        newSplit.test === 0;
+      const sumsToOne = checkSplit(
+        newSplit.train,
+        newSplit.validation,
+        newSplit.test,
+      );
+
+      if (hasZero) {
+        setRandomSplitErrorText("All splits must be greater than 0");
+        setRandomSplitError(true);
```
cristian-tamblay: I would let Validation or Test be 0, you can train with 100% data if you want, but train split can't be 0, something like:


```suggestion
      const hasZeroTrain = newSplit.train === 0;
      const sumsToOne = checkSplit(newSplit.train, newSplit.validation, newSplit.test);
      
      if (hasZeroTrain) {
        setRandomSplitErrorText("Train split must be greater than 0");
        setRandomSplitError(true);
```

I think huggingface backend allows this

cristian-tamblay (changes requested):
Small changes to allow 0 split at val/test

DashAI/front/src/components/experiments/SplitDatasetRows.jsx
Irozuku: if that's the case then should it be allowed to train with less than 100% of the dataset?. E.g.

- train = 0.6
- test = 0,2
- val = 0

cristian-tamblay (approved):

PR #365, opened 2025-10-30
Fix/dataset filter density buttons

Creylay:
#### Density selector was not working

---

#### Density in DataGrid (MUI X v7+)

**Problem:**

The `density` prop was being passed directly to the DataGrid component, which prevented users from changing the row density using the toolbar (GridToolbarDensitySelector). Additionally, with MUI X v7+, density must be controlled by the DataGrid’s internal state for the selector to work properly.

**Solution:**

- Removed the `density` prop from DataGrid.
- Set the default density using `initialState={{ density: 'compact' }}`.
- Now users can change the density from the toolbar and the change is correctly reflected in the table.

---

#### Dataset Table Filters Fix

**1. Correct filter model sent from frontend**

- The `DatasetTable.jsx` component uses the MUI DataGrid filter model (`filterModel`) and sends it as-is to the backend whenever the user applies a filter.
- The `fetchPage` method receives the filter model and passes it to `getDatasetFileFiltered`, which serializes and sends it to the `/v1/dataset/filter/` endpoint.
- Any transformation or overwriting of the operator (`operatorValue`) was removed, ensuring operators like `"equals"` and `"isAnyOf"` are sent correctly to the backend.

**2. Backend interprets and applies filters**

- The backend receives the `filterModel` and applies the filters to the Arrow table, using the received operator and value.
- Type conversion was fixed so that filter values match the column type (e.g., numbers, booleans, strings).
- Support was implemented for all relevant operators: `"contains"`, `"equals"`, `"isAnyOf"`, etc.

**3. Pagination and row count**

- The backend always returns the total number of rows after filters are applied, allowing the frontend to display the correct number of pages and rows.
- The frontend now always uses the total returned by the backend to show the correct number of pages and rows.


cristian-tamblay (approved):
Filters dont work in notebooks yet, but this works in datasets at least

PR #385, opened 2025-11-20
Updates migration with alembic 

Creylay:
## Summary

This PR redo the migrations configuration that were first applied here: https://github.com/DashAISoftware/DashAI/pull/303 and reverted here: https://github.com/DashAISoftware/DashAI/pull/329

This PR (re)introduce database migration using alembic (automatic migrations). The main objective is to be able to update the database without deleting the current one. Migrations scripts are generated using alembic at dev command and are applied automatically at app startup.

Two main features of this PR is that migrations are applied automatically on app startup and adds CI configuration to check database consistency between models and migrations scripts (to ensure there is no db drift).

### Note (previous bug of https://github.com/DashAISoftware/DashAI/pull/303)

It was reported that migrations weren't being applied in the packaged version of DashAI. i.e: after installing it with `pip install .`. This was caused because alembic configuration wasn't packaged and couldn't be load properly. This is fixed in this PR.

---

## Type of Change

- [x] Backend change
- [ ] Frontend change
- [ ] CI / Workflow change
- [ ] Bug fix
- [ ] Documentation

---

## Changes (by file)
- Update README.rst with alembic / migration instructions
- Update alembic configuration to run db migrations. Move alembic config files to `DashAI/alembic`
- Update `MANIFEST.in` to package alembic files.
- Reset migrations (delete old ones and create an initial one)
- Runs migrations at app startup. On `DashAI/back/app.py`
- Adds CI configuration to check consitency between models and migrations. This will ensure that all PR's that try to merge to develop ensures that migrations are in order (if they modify database.py then they should create the correspondent migration file).
---

## Testing (optional)

No tests added in `testing/` but a new CI configuration is added to check drift between migrations and models.

Tested running with:

```
python -m DashAI
```

And after 

```
pip install .

dashahi
```

And migrations were applied correctly on startup.


cristian-tamblay:
Will wait on executables to confirm this is compatible.

cristian-tamblay (approved):

PR #388, opened 2025-11-25
Fix: sanitize correlation matrix to prevent JSON non-finite value error

Creylay:
### Summary
This PR fixes a server error when serializing dataset info that contains non-finite float values.  
The code in `dashai_dataset.py` now sanitizes non-finite floats (NaN, +Inf, -Inf) before JSON serialization by converting them to `None` (JSON `null`).  
This prevents the error **"Out of range float values are not JSON compliant"** and ensures dataset info endpoints return valid JSON.

### Type of Change
Check all that apply:
- [x] Backend change
- [ ] Frontend change
- [ ] CI / Workflow change
- [x] Bug fix
- [ ] Documentation

### Changes (by file)

**dashai_dataset.py**
- Dropped correlation matrix rows and columns that are entirely NaN before building the output dictionary.
- Prevents non-finite values from reaching JSON serialization.

Correlation matrices can produce NaNs when columns have constant values or insufficient overlap. Cleaning NaN-only rows/columns ensures the /info endpoint returns valid JSON.

### DIFF
--- a/DashAI/back/dataloaders/classes/dashai_dataset.py
+++ b/DashAI/back/dataloaders/classes/dashai_dataset.py
@@ -2,7 +2,6 @@
 
 import json
 import logging
-import math
 import os
 import uuid
 from typing import Dict, List, Literal, Tuple, Union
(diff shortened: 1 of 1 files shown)


DashAI/back/dataloaders/classes/dashai_dataset.py
```diff
@@ -279,12 +290,27 @@ def compute_metadata(self) -> "DashAIDataset":
         # --- Correlations ---
         if not numeric_cols.empty:
             corr_matrix = numeric_cols.corr(numeric_only=True)
+            # Drop nan columns and rows from correlation matrix
+            corr_matrix = corr_matrix.dropna(axis=0, how="all").dropna(
+                axis=1, how="all"
+            )
```
Irozuku: This should be enough to cover the edge case where a column contains only zeros (resulting in zero variance, and therefore a division by zero when computing the correlation). The remaining changes add unnecessary checks, since the logic still works even with columns that contain NaN or null values.

Irozuku (changes requested):

Irozuku (approved):

cristian-tamblay (approved):

PR #443, opened 2026-02-01
fix reset step state when selecting dataset or notebook

Felipedino:
This pull request makes a small update to the dataset and notebook selection logic in the `DatasetsContent` component. Now, when a user selects a dataset or a notebook, the current step is reset to 0, ensuring the UI starts from the initial step for each new selection.

- User experience improvements:
  * Reset the `step` state to 0 when a dataset or notebook is selected, so the interface always starts from the beginning for each new selection.


Felipedino:
se arregló con el refactor de la Cami

PR #457, opened 2026-02-25
add FileMetadata extractor to tasks/utils

alekseevpavel04:
## Summary
Added FileMetadata dataclass and extract_file_metadata fn alongside the existing type detection helper. Callers can now get structured info (row/column counts for tabular formats, dimensions for images, file list for archives, encoding for text) in one call instead of just a type string.

---

## Type of Change

- [x] Backend change
- [ ] Frontend change
- [ ] CI / Workflow change
- [ ] Bug fix
- [ ] Documentation

---

## Changes (by file)
Briefly list the important modified files and what was done.

- `DashAI/back/tasks/utils.py`: added FileMetadata dataclass, five private per-type extractor fns (text, csv, json, image, archive), and the public extract_file_metadata fn that ties them together.

---

## Notes (optional)
PIL is imported lazily inside the image extractor so it stays optional. CSV detection uses csv.Sniffer as a fallback for plain text inputs that turn out to be tabular. Archive listing is capped at 50 entries to avoid huge payloads.

### DIFF
--- a/DashAI/back/tasks/utils.py
+++ b/DashAI/back/tasks/utils.py
@@ -1,4 +1,14 @@
+import csv
+import io
+import json
+import logging
+import zipfile
+from dataclasses import dataclass, field
+from typing import Optional
+
 import filetype
+
+log = logging.getLogger(__name__)
 
 
 def get_bytes_with_type_filetype(data: bytes) -> tuple[bytes, str]:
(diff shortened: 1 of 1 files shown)


cristian-tamblay (approved):
Thank you, this might be useful in the future when we have image classification task as baseline for the software

PR #472, opened 2026-03-09
Revert transformers path

Irozuku:
## Summary
Reverts a previous pull request that **removed temporary checkpoint handling during model training**. This restores the behavior where models can use a dedicated temporary directory for Hugging Face training checkpoints and ensures the directory exists during the build process.

---

## Type of Change

- [x] Backend change
- [ ] Frontend change
- [x] CI / Workflow change
- [ ] Bug fix
- [ ] Documentation

---

## Changes (by file)
- `.github/workflows/build.yml`: Restores creation and inclusion of the `DashAI/back/user_models/temp_checkpoints` directory in PyInstaller build steps for all platforms.
- `DashAI/back/models/.../distilbert_model.py`: Restores the use of a temporary checkpoint directory for the Hugging Face `Trainer` and cleanup of the directory after training.
- `DashAI/back/models/.../opus_mt_en_es_transformer.py`: Reintroduces the temporary checkpoint directory for training and removes it after training completes.

---

## Testing (optional)
- Verify that the build workflows correctly create and package the `temp_checkpoints` directory.
- Confirm that model training runs successfully and that temporary checkpoints are cleaned up after training.

---

## Notes (optional)
This PR **reverts the removal of temporary checkpoint handling**, restoring support for models that require temporary storage during training while preventing leftover checkpoint files.


cristian-tamblay (approved):

cristian-tamblay:
I dont know if there's a way to parametrize this so windows+mac executables use the same flags (or at least some), because im sure we will fail adding these to both of them in the future

cristian-tamblay (changes requested):
Fix conflicts and maybe try to parametrize flags?

dashai.spec
```diff
@@ -0,0 +1,58 @@
...
+
+coll = COLLECT(
+    exe,
+    a.binaries,
+    a.datas,
+    strip=False,
+    upx=False,
+    upx_exclude=[],
+    name="DashAI-launcher-cpu",
+)
```
Creylay, line 58: On macOS, the COLLECT step only creates a folder with binaries.  The BUNDLE step is required to wrap that folder into a valid '.app' package  with the proper structure (Contents/MacOS/...) and metadata.

```suggestion
)

if platform.system() == "Darwin":
    app = BUNDLE(
        coll,
        name='DashAI.app',
        icon=None, 
        bundle_identifier='com.dashai.app',
        info_plist={
            'NSHighResolutionCapable': 'True',
            'LSBackgroundOnly': 'False',
        },
    )
```

Creylay (changes requested):

cristian-tamblay (approved):

PR #481, opened 2026-03-13
refactor: centralize dataset polling and remove redundant fetchDatase…

Creylay:
## Summary
Refactored dataset status updates to use a single centralized polling flow instead of component-level polling triggers. This removes duplicated polling behavior when datasets are processing and keeps both the left dataset list and `DatasetVisualization` in sync from shared state.

---

## Type of Change

- [ ] Backend change
- [x] Frontend change
- [ ] CI / Workflow change
- [x] Bug fix
- [ ] Documentation

---

## Changes (by file)
- `DashAI/front/src/hooks/datasets/useDatasets.js`:
  - Centralized dataset refresh behavior using job updates (`subscribeJobs`)
  - Made `fetchDatasets` stable and based on a ref-backed latest dataset snapshot
  - Removed extra `fetchDatasets()` call from `startDatasetPolling` success callback (now selection/snackbar only)
- `DashAI/front/src/components/DatasetVisualization.jsx`:
  - Removed local polling trigger (`setTimeout(fetchDatasets, 1000)`)
  - Simplified effect to react to dataset changes and fetch dataset info only
  - Removed unused `fetchDatasets` prop
- `DashAI/front/src/components/notebooks/dataset/DatasetsCenterContent.jsx`:
  - Removed passing `fetchDatasets` prop into `DatasetVisualization`
- `DashAI/front/src/utils/jobPoller.js`:
  - Removed individual polling interval from `startJobPolling` (eliminated race condition)
  - Now relies exclusively on global polling system via `pollJobs()` mechanism
  - Prevents duplicate callbacks being triggered simultaneously (fixes duplicate success messages)

---

## Testing (optional)
1. Start uploading/processing a dataset and open the dataset module.
2. Verify dataset status updates in the left list while processing.
3. Verify `DatasetVisualization` updates correctly when processing finishes.
4. Confirm no duplicated polling behavior (fewer duplicate refreshes/requests).


DashAI/front/src/components/DatasetVisualization.jsx
```diff
@@ -52,30 +51,11 @@ export default function DatasetVisualization({
...
     const fetchDatasetInfo = async () => {
-      const isProcessing = !(dataset.status === 3 || dataset.status === 4);
-      if (isProcessing && fetchDatasets) {
-        setTimeout(() => {
-          fetchDatasets();
-        }, 1000);
-        return;
-      }
       try {
         const info = await getDatasetInfo(Number(dataset.id));
```
Irozuku, line 90: getDatasetInfo is being polled unnecessarily while the dataset is being uploaded. Not only here but also at the 

Image

DashAI/front/src/hooks/datasets/useDatasets.js
```diff
@@ -58,9 +63,31 @@ export function useDatasets({ t }) {
 
   const fetchDatasets = useCallback(async () => {
     const data = await getDatasets();
-    const enriched = await enrichDatasetsWithInfo(data, datasets);
+    const enriched = await enrichDatasetsWithInfo(data, datasetsRef.current);
```
Irozuku, line 58: This also calls `getDatasetInfo` without needing to do so.

Image

Irozuku (changes requested):

DashAI/front/src/components/DatasetVisualization.jsx
Creylay: Fixed. I've added a conditional guard in the useEffect to ensure getDatasetInfo is only called when the dataset status is strictly 'Finished' (status 3). This prevents premature requests and 422 errors while the dataset is still processing or uploading. Additionally, I added a try/catch block to handle cases where the info might not be immediately available.

DashAI/front/src/hooks/datasets/useDatasets.js
Creylay: Optimized the enrichDatasetsWithInfo helper to prevent redundant API calls. It now uses a ref-backed comparison to check if the dataset info (rows/columns) is already present in the local state. If the data exists or the dataset is not in a 'Finished' state, the API call is skipped. This ensures that we only fetch info for newly processed datasets.

Irozuku (dismissed):

Irozuku (approved):

PR #653, opened 2026-05-28
Add image classification models

Creylay:
## Summary

Adds four new torchvision-based image classification models to DashAI and introduces a shared base class that consolidates all training logic. The MLP image classifier is also upgraded with new configurable hyperparameters. A defensive logging improvement in `BaseModel` avoids silent failures when a metric returns a non-finite score (e.g., only one class present in a split).

  ---

  ## Type of Change

  - [x] Backend change
  - [ ] Frontend change
  - [ ] CI / Workflow change
  - [ ] Bug fix
  - [ ] Documentation

  ---

  ## Changes (by file)

  - `DashAI/back/models/base_torchvision_image_classifier.py`: New shared base class (`TorchvisionImageClassifier`) for torchvision-backed models. Implements the full train/predict/save/load lifecycle, data augmentation pipeline, and a shared schema (`TorchvisionImageClassifierSchema`) with parameters: `epochs`, `learning_rate`, `batch_size`, `image_size`, `dropout_rate`, `weight_decay`.
  - `DashAI/back/models/cnn_image_classifier.py`: New `CNNImageClassifier` — a lightweight custom CNN with two conv blocks and a configurable number of filters.
  - `DashAI/back/models/lenet5_image_classifier.py`: New `LeNet5ImageClassifier` — LeNet-5 architecture (LeCun et al., 1998), adapted for multi-class output.
  - `DashAI/back/models/resnet18_image_classifier.py`: New `ResNet18ImageClassifier` — ResNet-18 backbone with optional ImageNet pre-trained weights; final FC layer replaced to match target classes.
  - `DashAI/back/models/resnet50_image_classifier.py`: New `ResNet50ImageClassifier` — ResNet-50 backbone with optional ImageNet pre-trained weights; final FC layer replaced.
  - `DashAI/back/models/efficientnet_b0_image_classifier.py`: New `EfficientNetB0ImageClassifier` — EfficientNet-B0 backbone with optional ImageNet
  - `DashAI/back/models/mlp_image_classifier.py`: Added `batch_size`, `image_size`, `dropout_rate`, and `weight_decay` parameters; wired them through training, prediction, save, and load paths.
  - `DashAI/back/models/base_model.py`: Added warning log when a metric returns a non-finite value (e.g., `NaN`, `inf`) during evaluation; the metric is skipped instead of crashing or silently producing bad results.
  - `DashAI/back/tasks/image_classification_task.py`: Shortened the `DESCRIPTION` field to a concise single sentence.
  - `DashAI/back/initial_components.py`: Registered `CNNImageClassifier`, `LeNet5ImageClassifier`, `ResNet18ImageClassifier`, `ResNet50ImageClassifier`, and `EfficientNetB0ImageClassifier`. Removed `VGG16ImageClassifier` (also deleted its file).
    
  ---
  
  ## Notes

  - All new models subclass `TorchvisionImageClassifier` and only need to implement `_build_backbone` and `_classifier_head`, keeping each model file minimal.
  - Pre-trained weights are downloaded from torchvision on first use when `pretrained=True` (the default).


Irozuku (changes requested):
This is missing lazy imports for heavy dependencies such as `torch`. These imports should be moved inside the methods where they are actually used to avoid unnecessary startup overhead and memory consumption when the functionality is not needed.

It is also missing live metrics reporting for classification models. A call to `calculate_metrics(...)` should be added in each model so metrics are computed and reported consistently during training and evaluation.

Irozuku (changes requested):
All models are missing a `device` parameter to explicitly select the execution device.

cristian-tamblay:
Also missing de language (and i think this show we are also missing pt language on models)

cristian-tamblay (approved):

PR #689, opened 2026-06-07
fix: update file size calculation in dataset info

Felipedino:
This pull request updates how dataset file size is calculated and displayed in the dataset visualization component. Instead of reporting the in-memory size, the system now reports the actual file size based on the Arrow table, ensuring more accurate information is shown to users.

**Backend: File size calculation update**
- Changed the backend to report `file_size_mb` using the Arrow table's byte size instead of the DataFrame's memory usage, providing a more accurate file size metric. (`DashAI/back/dataloaders/classes/dashai_dataset.py`)

**Frontend: File size display update**
- Updated the frontend to use the new `file_size_mb` field for displaying file size in the dataset visualization header, instead of the previous `memory_usage_mb` field. (`DashAI/front/src/components/DatasetVisualization.jsx`)

### DIFF
--- a/DashAI/back/dataloaders/classes/dashai_dataset.py
+++ b/DashAI/back/dataloaders/classes/dashai_dataset.py
@@ -881,6 +881,9 @@ def save_dataset(
         }
     )
 
+    if "general_info" in metadata:
+        metadata["general_info"]["file_size_mb"] = os.path.getsize(data_filepath) / 1e6
+
     with open(metadata_filepath, "w", encoding="utf-8") as f:
         json.dump(metadata, f, indent=2, sort_keys=True, ensure_ascii=False)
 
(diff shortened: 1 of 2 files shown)


cristian-tamblay:
I find this change a bit weird. Any comments @Irozuku or @Creylay ?

Felipedino:
It previously used the size of the pandas DataFrame. Now, it retrieves the size from the Arrow file generated in .dashai

DashAI/back/dataloaders/classes/dashai_dataset.py
```diff
@@ -214,7 +214,7 @@ def _compute_general_info(self, dataset_df) -> dict:
         return {
             "n_rows": len(dataset_df),
             "n_columns": len(dataset_df.columns),
-            "memory_usage_mb": float(dataset_df.memory_usage(deep=True).sum() / 1e6),
+            "file_size_mb": float(self.arrow_table.nbytes / 1e6),
```
Irozuku: - Changing the key carries unnecesary changes, I recommend mantaining the same `memory_usage_mb` key to avoid this.
- `arrow_table.nbytes` returns the amount of bytes used in memory which can differ slightly from the real memory used in disk.

Irozuku (changes requested):

cristian-tamblay (approved):

PR #707, opened 2026-06-18
Add a Linux AppImage to the release builds

Irozuku:
## Summary
Adds a Linux AppImage build to the release workflow so every release ships a self contained, CPU only Linux executable alongside the existing Windows/macOS installers. The AppImage is built with `python-appimage` (embeds a manylinux CPython + the app and its deps), so end users need no Python installed. Includes small launcher fixes required to make the app run correctly inside an AppImage.

---

## Type of Change

- [x] Backend change
- [ ] Frontend change
- [x] CI / Workflow change
- [x] Bug fix
- [ ] Documentation

---

## Changes (by file)
- `.github/workflows/publish.yml`: new `build-linux-appimage` job (runs on `ubuntu-22.04` for glibc 2.35). Builds the wheel (frontend bundled), generates the icon, builds the AppImage via `python-appimage`, and uploads it. The `github-release` job now depends on it and publishes `dashAI-<version>-x64-linux.AppImage`. CPU torch/llama-cpp wheels are selected via `PIP_INDEX_URL`/`PIP_EXTRA_INDEX_URL`.
- `appimage/requirements.txt`: recipe deps (bare names only — `torch`, `torchvision`, `llama-cpp-python`); the locally built wheel path is appended at build time.
- `appimage/entrypoint.sh`: runs the `dashai` console script via the embedded Python; relaunches inside a terminal emulator when double-clicked (headless fallback if none).
- `appimage/dashai.desktop`, `appimage/dashai.appdata.xml`: desktop + AppStream metadata required by the recipe.
- `appimage/.gitattributes`: forces LF endings on the shell/desktop files.
- `DashAI/__main__.py`: run the Huey consumer in a thread when running inside an AppImage (detected via `APPDIR`/`APPIMAGE`), matching the existing PyInstaller `sys.frozen` path.

---

## Testing (optional)
- Built and ran the AppImage on a Debian VM with no system Python: server starts, Huey queue processes jobs, browser opens at `localhost:8000`, and CPU llama-cpp LLM inference works.

---

## Notes (optional)
- AppImage is CPU only, consistent with the Windows/macOS installers. GPU stays on the CUDA Docker image.
- Host requirements: glibc >= 2.35 and FUSE2 (or run with `--appimage-extract-and-run`). No Python needed.
- The double click terminal popup needs a terminal emulator installed; otherwise it runs headless and still opens the browser.

### DIFF
--- a/appimage/requirements.txt
+++ b/appimage/requirements.txt
@@ -1,0 +1,3 @@
+torch
+torchvision
+llama-cpp-python
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -230,7 +230,61 @@ jobs:
           retention-days: 1
 
   # ============================================================
-  # 6. Publish installers to GitHub Release
+  # 6. Build Linux AppImage
+  # ============================================================
+  build-linux-appimage:
+    needs: react-build
+    # ubuntu-22.04 sets the minimum glibc (2.35) for the AppImage.
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v4
+      - name: Download React build artifact
+        uses: actions/download-artifact@v4
+        with:
+          name: react-build
+          path: DashAI/front/build
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+      - name: Install build tooling
+        run: |
+          python -m pip install --upgrade pip
+          pip install build python-appimage
+          sudo apt-get update
+          sudo apt-get install -y libfuse2 imagemagick
+      - name: Build wheel (frontend bundled)
+        run: python -m build --wheel
+      - name: Verify frontend is included in wheel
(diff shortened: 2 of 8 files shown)


appimage/requirements.txt
```diff
@@ -0,0 +1,3 @@
+torch
+torchvision
+llama-cpp-python
```
cristian-tamblay, line 3: Can we pin these so they match the requirements-cpu.txt ? O that breaks the appimage with the extra indexes?

Irozuku: That's what I tried first, but it broke the AppImage build pipeline. Also, `llama-cpp-python` requires a specific installation process, so it isn't included in `requirements.txt` and therefore can't be used as a single source for this case.

cristian-tamblay (approved):

PR #756, opened 2026-07-03
Migration to uv, added instructions and fixed workflows

cristian-tamblay:
## Summary

Migrates Python packaging and dependency management from `setup.py` + `requirements*.txt` to a standard PEP 621 `pyproject.toml` managed with [uv](https://docs.astral.sh/uv/), adding a `uv.lock` for reproducible installs across dev, CI, and Docker.

Key points:
- **End users are unaffected**: the published wheel is standard, so `pip install dashai` keeps working exactly as before.
- The CPU/CUDA requirements variants are now extras: `uv sync --extra cpu` / `--extra cuda` pick the right PyTorch index automatically via `[tool.uv.sources]` — this solves the torch-index problem that made us abandon `pip-compile` in the past (see the old note in `requirements.txt`).
- CI now installs from the lockfile (`uv sync --locked`), so builds are reproducible and dependency caching is handled by `setup-uv`.
- Fixes a packaging bug: the published wheel included the whole `tests/` package (`find_packages()` had no filter); packages are now scoped to `DashAI*`.

---

## Type of Change

- [ ] Backend change
- [ ] Frontend change
- [x] CI / Workflow change
- [x] Bug fix
- [x] Documentation

---

## Changes (by file)
- `pyproject.toml`: full PEP 621 metadata (moved from `setup.p `requirements*.txt`), `cpu`/`cuda` extras with per-extraPyTorch/llama-cpp indexes, dev dependency group. Setuptools reANIFEST.in` keeps bundling the frontend.
- `uv.lock` (new): universal lockfile (288 packages, all platf
- `setup.py`, `requirements{,-dev,-cpu,-cuda,-plugins}.txt`: d.txt` was unreferenced and pinned torch 1.13 from 2022).
- `.github/workflows/build-test.yaml`, `db-migrations.yaml`: `` + `uv run ...`.- `.github/workflows/publish.yml`: `uv build` + `uvx twine`; P--extra cpu --no-dev` and `uv run --no-sync`; release version isnow extracted from `pyproject.toml`.- `.github/workflows/docs.yaml`, `pre-commit.yaml`: install inpip install --system` (the Docusaurus build and a `language:system` hook invoke `python` directly).
- `Dockerfile`, `Dockerfile.cuda`: install with `uv sync --locDA image builds llama-cpp with `CMAKE_ARGS="-DGGML_CUDA=on"` as
before.
- `update-server.sh`: uses uv when available, falls back to pi
- `README.rst`, `docs/**/dev-setup.md` (en/es), `CLAUDE.md`, `uv is the recommended path (including `uv tool install dashai`),
pip instructions kept.

---

## Testing

- Full backend suite passes with the new environment: `uv run pytest tests/` → 541 passed.                                                             - Wheel built with `uv build` passes `twine check`; verified i`, `DashAI/alembic`, seeds and static images, and no longerships `tests/`.
- Smoke test: server boots from the uv-managed venv and `/api/mponent registry.
- `uv sync --locked --extra cpu` and `--extra cuda` resolve coend locally; the CUDA llama-cpp compilation is only exercised by
`Dockerfile.cuda`, same as before).

---

## Notes

- **pip survives on purpose in three places** — please don't " installer (`DashAI/back/plugins/utils.py`) runs in the enduser's env where pip is guaranteed; the AppImage job uses pythupdate-server.sh` falls back to pip if the server lacks uv.
- The PyTorch index selection in the extras lives in `[tool.uval and not published to PyPI — pip users still follow the manual
`--index-url` instructions in the README (unchanged).
- Gotcha for contributors: `uv run` re-syncs the env to the de`uv sync --extra cpu`, use `uv run --no-sync` or re-run thesync.
- Contributors now need uv installed (one-liner, see README); e watched on its first run against `production`.


docs/docs/build/dev-setup.md
```diff
@@ -22,22 +22,32 @@ git checkout develop
...
 
 ```bash
-pip install -r requirements.txt
-pip install -e .
-pip install -r requirements-dev.txt
+uv sync --extra cpu
+```
+
+Alternatively, plain `pip` works inside any environment (venv or conda),
+since all metadata lives in `pyproject.toml` — note this skips the lockfile,
```
Irozuku: remove em dash

docs/i18n/es/docusaurus-plugin-content-docs/current/build/dev-setup.md
```diff
@@ -22,22 +22,33 @@ git checkout develop
...
 
 ```bash
-pip install -r requirements.txt
-pip install -e .
-pip install -r requirements-dev.txt
+uv sync --extra cpu
+```
+
+Como alternativa, `pip` a secas funciona dentro de cualquier entorno (venv o
+conda), ya que toda la metadata vive en `pyproject.toml` — ojo que esto no usa
```
Irozuku: remove em dash

pyproject.toml
```diff
@@ -1,3 +1,136 @@
+[build-system]
+requires = ["setuptools>=77"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "DashAI"
```
Irozuku: dashAI

pyproject.toml
```diff
@@ -1,3 +1,136 @@
...
+version = "0.9.6"
+description = "DashAI: a graphical toolbox for training, evaluating and deploying state-of-the-art AI models."
```
Irozuku: dashAI

pyproject.toml
```diff
@@ -1,3 +1,136 @@
...
+readme = "README.rst"
+license = "MIT"
+authors = [{ name = "DashAI Team", email = "fbravo@dcc.uchile.cl" }]
```
Irozuku: dashAI

README.rst
```diff
@@ -113,6 +136,9 @@ With the environment active:
 This step is optional on CPU (step 2 already installed a working PyTorch).
 Run the section below that matches your hardware to pick a specific build.
 
+Every ``pip install`` command below can also be run as ``uv pip install`` with
+the same flags — same result, just faster.
```
Irozuku: remove em dash

DashAI/back/dataset_sources/openml_dataset_source.py
```diff
@@ -197,7 +197,7 @@ def _meta(did: str) -> tuple[str, tuple[str, ...]] | None:
                 metas = list(pool.map(_meta, ids))
 
             entries = []
-            for row, did, meta in zip(rows, ids, metas):
+            for row, did, meta in zip(rows, ids, metas, strict=False):
```
Irozuku, line 200: This seems out of scope

Irozuku (changes requested):

Irozuku:
> * `uv sync --locked --extra cpu` and `--extra cuda` resolve coend locally; the CUDA llama-cpp compilation is only exercised by

why does `--extra cuda` only install the wheels for torch?. Also it seems to force the installation of cpu version when running the backend:

```
C:\Users\emili\Desktop\dash\DashAI> uv sync --locked --extra cuda
Resolved 288 packages in 1ms
Installed 1 package in 6.13s
 + torch==2.11.0+cu128
C:\Users\emili\Desktop\dash\DashAI> uv run python -m DashAI -nb                      
Uninstalled 2 packages in 771ms
Installed 2 packages in 7.19s
```

Checking inside the platform, GPU's are not being recognized.

DashAI/back/dataset_sources/openml_dataset_source.py
cristian-tamblay: This PR adds requires-python = ">=3.10" to pyproject.toml (the [project] table didn't exist before). Ruff derives its target-version from that field, and with a 3.10+ target the B905 rule (flake8-bugbear) becomes active, which requires an explicit strict= in zip(), a parameter that only exists since Python 3.10. Since the pre-commit CI runs with --all-files, without this fix the pipeline of this very PR goes red because of a pre-existing zip().

cristian-tamblay:
> > * `uv sync --locked --extra cpu` and `--extra cuda` resolve coend locally; the CUDA llama-cpp compilation is only exercised by
> 
> why does `--extra cuda` only install the wheels for torch?. Also it seems to force the installation of cpu version when running the backend:
> 
> ```
> C:\Users\emili\Desktop\dash\DashAI> uv sync --locked --extra cuda
> Resolved 288 packages in 1ms
> Installed 1 package in 6.13s
>  + torch==2.11.0+cu128
> C:\Users\emili\Desktop\dash\DashAI> uv run python -m DashAI -nb                      
> Uninstalled 2 packages in 771ms
> Installed 2 packages in 7.19s
> ```
> 
> Checking inside the platform, GPU's are not being recognized.

 Why GPUs stopped being recognized: uv run without flags re-syncs the environment to the default dependency set (no extras). That is the Uninstalled 2 / Installed 2 you saw: it swapped your torch==2.11.0+cu128 back to the default PyPI build, which on Windows is CPU-only. Pass the same extra you synced with:
 
uv sync --locked --extra cuda
uv run --extra cuda python -m DashAI -nb

1. (uv run --no-sync ... also works.) We added a note about this to the README, the dev-setup docs and CLAUDE.md, since it's an easy trap.

2. Why --extra cuda "only installs torch": torch and torchvision are the only pieces with prebuilt CUDA wheels (from the cu128 index, Windows included). llama-cpp-python has no prebuilt CUDA wheel on PyPI (sdist only), so it compiles on your machine and only gets CUDA offload if you set CMAKE_ARGS="-DGGML_CUDA=on" with CMake + MSVC + the CUDA toolkit installed (see the "Build tools for GPU llama-cpp" section in the README; Dockerfile.cuda does exactly this). This is the same situation the old requirements-cuda.txt had. Also, "Installed 1 package" just means the incremental sync only changed what differed from your previous env.

Irozuku:
Actually, the old `requirements-cuda.txt` configured the environment variables and installed both `torch` and `llama-cpp-python` (as long as everything was configured properly). So, as far as I can tell, to install `llama-cpp-python` I would need to do it manually (i.e., set the environment variables and then run `uv pip install llama-cpp-python`)?

cristian-tamblay:
> Actually, the old `requirements-cuda.txt` configured the environment variables and installed both `torch` and `llama-cpp-python` (as long as everything was configured properly). So, as far as I can tell, to install `llama-cpp-python` I would need to do it manually (i.e., set the environment variables and then run `uv pip install llama-cpp-python`)?

Almost: the cuda extra does install llama-cpp-python automatically, so there is no manual uv pip install involved. uv sync --extra cuda installs torch (cu128 wheels) + torchvision + llama-cpp-python, same package set as the old requirements-cuda.txt.

What did move to the command line is only the CMake flag. Requirements files could carry per-package config settings inline (llama-cpp-python -C cmake.args="-DGGML_CUDA=on"), but PEP 621 extras cannot express build flags, so it is passed as an environment variable in front of the same single command:

CMAKE_ARGS="-DGGML_CUDA=on" uv sync --extra cuda --reinstall-package llama-cpp-python

I tested this and I had the option for GPU on llms

Irozuku (approved):

PR #776, opened 2026-07-27
feat: enhance search functionality to support substring matching and smote exclude target column

Felipedino:
This pull request improves the search functionality in the `RightBar` component by making the search more flexible and accurate. Instead of matching the entire search query as a single string, the search now splits the query into individual tokens and checks if all tokens are present in the item's display name or description.

**Search improvements:**

* The search query is now split into tokens, and the filtering logic checks that all tokens are present in the display name or description, allowing for more flexible multi-word searches. [[1]](diffhunk://#diff-711647abd9237f7984867e4544ede07fcf00a688129ce0f7ae4dd2c752781c81R315) [[2]](diffhunk://#diff-711647abd9237f7984867e4544ede07fcf00a688129ce0f7ae4dd2c752781c81L327-R335)
* The filter logic now returns all items if there are no tokens (i.e., the search box is empty), maintaining previous behavior for empty queries.

### DIFF
--- a/DashAI/back/converters/imbalanced_learn/smote_converter.py
+++ b/DashAI/back/converters/imbalanced_learn/smote_converter.py
@@ -14,22 +14,8 @@ from DashAI.back.core.schema_fields.base_schema import BaseSchema
 from DashAI.back.core.utils import MultilingualString
 from DashAI.back.types.categorical import Categorical
 from DashAI.back.types.dashai_data_type import DashAIDataType
+from DashAI.back.types.utils import NON_NUMERIC_DTYPES
 from DashAI.back.types.value_types import Float, Integer
-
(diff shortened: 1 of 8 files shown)


DashAI/back/converters/imbalanced_learn/smote_converter.py
```diff
@@ -12,9 +12,25 @@
...
+    "int64",
+    "uint8",
+    "uint16",
+    "uint32",
+    "uint64",
+    "float16",
+    "float32",
+    "float64",
+    "bool",
+]
```
cristian-tamblay: This is only for this converter? Can't we have this in the base converter class?

cristian-tamblay (approved):

PR #784, opened 2026-07-30
Fix models session redundant requests

Creylay:
## Summary
Fixes a cluster of redundant/duplicate network requests in the Models module, found while auditing backend logs during normal use (create session → add model → train → view results). Multiple components independently fetched the same data (component lists, model schemas, session details) instead of sharing it, and a WebSocket connection was being torn down and reopened unnecessarily. Also changes the "no runs yet" case from a 404 to an empty 200 response, since it's a normal state, not an error.

---

## Type of Change

- [x] Backend change
- [x] Frontend change
- [ ] CI / Workflow change
- [x] Bug fix
- [ ] Documentation

---

## Changes (by file)

- `DashAI/back/api/api_v1/endpoints/runs.py`: `GET /run/?model_session_id=` no longer raises 404 when the session exists but has no runs yet, returns `200` with an empty list. 404 is now reserved for when the session itself doesn't exist.
- `DashAI/front/src/hooks/models/useSessions.js`: `fetchRuns` no longer special-cases 404 as "no runs" (that case is now a normal `200 []`), so a real 404 (session not found) surfaces as an actual error instead of being silently swallowed.
- `DashAI/front/src/hooks/models/useModelComponents.js` (new): shared cache for the `Model`/`Metric` component lists. Fetched once and reused across consumers instead of each
one fetching independently. Includes an unfiltered list plus a per-task-namuest de-duplication.
- `DashAI/front/src/components/models/ModelsContext.jsx`: wires up `useModelComponents` and exposes `allModels`, `allMetrics`, and `getModelsForTask` to the rest of the module.
- `DashAI/front/src/components/models/SessionVisualization.jsx`: consumes `stead of fetching its own copy on mount.
- `DashAI/front/src/components/models/ModelComparisonTable.jsx`: consumes `allModels`/`allMetrics` from context instead of independently re-fetching both lists every time it mounts.
- `DashAI/front/src/components/models/ModelsRightBar.jsx`: uses `getModelsForTask(session.task_name)` instead of calling the API directly, so switching between sessions with the same task no longer re-fetches the same model list.
- `DashAI/front/src/pages/results/components/ResultsGraphs.jsx`: accepts an optional `metrics` prop to reuse a caller's already-fetched list; falls back to its own fetch when not provided, so it still works standalone outside the Models module.
- `DashAI/front/src/hooks/useSchema.js`: added in-flight-promise de-duplica consumers of the same model's schema (e.g. a dialog's own prefetch and theform it renders a step later) share one request instead of each firing its own.
- `DashAI/front/src/components/models/LiveMetricsChart.jsx`: split the WebSocket effect in two. The socket now only reconnects when `run.id` changes (not on every `run.status` change, which caused unnecessary reconnects when a run finished); the "resestart" logic moved to its own effect. Also removed its own`getModelSessionById` fetch in favor of the `modelSessionDetail` passed down from the parent.
- `DashAI/front/src/components/models/runResults/useRunResultsData.js`: exposes the already-fetched `modelSessionDetail` so sibling components can reuse it instead of re-fetching.
- `DashAI/front/src/components/models/RunResults.jsx`: passes `modelSessionDetail` down to `LiveMetricsChart`.

---

## Testing (optional)
- Create a session, add a model, and train it while watching the Network tab: the model's component schema (`/component/{name}/`), `/model-session/{id}`, and the Model/Metric component lists should each be requested once instead of 2-3 times.
- Confirm the live-metrics WebSocket (`/metrics/ws/{run_id}`) opens once pect when the run transitions from running to finished.
- Switch between two sessions that share the same task and confirm the available-models list isn't re-fetched.
- Navigate to a session with zero runs and confirm the run list request ret404`.


DashAI/front/src/components/models/LiveMetricsChart.jsx
```diff
@@ -150,27 +150,19 @@ export function LiveMetricsChart({ run }) {
         console.log("WebSocket already closed");
       }
     };
-  }, [run.id, run.status]);
+  }, [run.id]);
```
cristian-tamblay: This breaks websocket

cristian-tamblay (changes requested):

DashAI/front/src/components/models/LiveMetricsChart.jsx
Creylay: Done

cristian-tamblay (approved):

PR #803, opened 2026-08-10
Feat/explainer story i18n

Creylay:
## Summary

Adds a `story()` narrative hook to every explainer (13/13), so global and local explanations come with a deterministic, human-readable summary in addition to their plots. Stories are generated in all 5 supported languages (`en`, `es`, `pt`, `de`, `zh`) at once and returned alongside the plot response, so the frontend can switch languages instantly (via the existing i18n selector) without refetching. Also removes redundant hardcoded English text that a subset of explainers previously baked into their plots, now superseded by the multilingual `story()` output.

---

## Type of Change

- [ ] Backend change
- [x] Backend change
- [x] Frontend change
- [ ] CI / Workflow change
- [x] Bug fix
- [ ] Documentation

---

## Changes (by file)

**Core story infrastructure**
- `DashAI/back/explainability/story.py`: new `format_story()`/`concat_stories()` helpers that build a `MultilingualString` from per-language templates, reused by every explainer instead of duplicating 5-language boilerplate.
- `DashAI/back/explainability/global_explainer.py`, `local_explainer.py`: added an optional `story(explanation, explainer_output)` hook to `BaseGlobalExplainer`/`BaseLocalExplainer` (default `None`), mirroring the existing `plot()` contract.

**`DashAI/back/explainability/explainers/ (Explainers)`**
- `permutation_feature_importance.py`, `regression_permutation_feature_importance.py`: new `story()` ranking features by importance; only claims the model "relies on" features with measurable (>0) importance, calling out separately when the rest showed none instead of lumping them together.
- `partial_dependence.py`, `regression_partial_dependence.py`: new `story()` classifying each curve as increasing/decreasing/flat/non-monotonic from its actual values; added a dedicated "flat" case for curves with no real change (previously misreported as "increases").
- `kernel_shap.py`: new `story()`; removed the old English-only Plotly annotation baked into the chart (predicted class/probability), now covered — with more detail — by `story()`.
- `contrastive_shap.py`, `dice_counterfactual.py`, `nearest_counterfactual.py`, `regression_kernel_shap.py`, `token_ablation.py`, `grad_cam.py`, `occlusion_saliency.py`, `lime_text.py`: migrated each explainer's existing hardcoded English `TextArtifact` summary into a multilingual `story()` with equivalent wording, and removed the now-redundant `TextArtifact` from `plot()`.

** `DashAI/back/api/api_v1/endpoints/explainers.py` (API)**
- `GET /global/plot/{id}` and `GET /local/plot/{id}` now also load `explanation_path` (previously only `plot_path`/`plots_path`) and attach a `"story"` key (`{"en": ..., "es": ..., ...}` or `null`) to each artifact/group in the response.
- Added `_resolve_story_explainer()` to reconstruct the explainer from its registered class + stored parameters (no trained model needed — `story()` never touches it) and `_attach_stories()`/`_attach_one_story()` to compute and attach stories per artifact, never raising: any failure (unbuildable explainer, mismatched shape) just leaves `"story": null`.
- Added `_as_group_target()`/`_as_artifact_target()` to reconstruct a real (unvalidated) `ArtifactGroup`/`Artifact` from the wire-format dict that `explainer_job.py` actually persists (it normalizes `plot()`'s output before pickling), so `story()`'s `isinstance` checks work against real data, not just hand-built test objects.

**`DashAI/front/src/components/explainers/` (Frontend)**
- `StoryBox.jsx`: new component that reads the active i18next language from `story` and renders it as a plain "text" artifact through the existing `ArtifactViewer` — same box, same download button, nothing new to maintain visually.
- `ExplainersPlot.jsx`: renders `StoryBox` under each grouped and ungrouped artifact.
- `utils/i18n/locales/{en,es,pt,de,zh}/explainers.json`: added the `storyTitle` label.

**Tests**
- `tests/back/api/test_explainer_story_attach.py` (new): covers `_attach_stories` against the real dict-shaped output `explainer_job.py` persists (not hand-built Pydantic objects), including the no-explainer no-op case.
- `tests/back/explainers/test_image_explainers.py`, `test_lib_explainers.py`, `test_new_explainers.py`, `test_task_explainers.py`: updated assertions that expected the now-removed `TextArtifact` in `plot()`'s output; assert `story()`'s content directly instead.

---

## Testing

- Manually verified against real generated explanations (not just synthetic test data) for both global and local explainers, in multiple languages, including the flat-curve and zero-importance edge cases.
- Reviewers should spot-check the wording of the non-English translations (es/pt/de/zh) in at least one explainer of each kind (global ranking, local per-instance, counterfactual). They were hand-translated, not machine-translated.

### DIFF
--- a/DashAI/back/api/api_v1/endpoints/explainers.py
+++ b/DashAI/back/api/api_v1/endpoints/explainers.py
@@ -1,4 +1,5 @@
 import logging
+from dataclasses import asdict
 from typing import TYPE_CHECKING
 
 from fastapi import APIRouter, Depends, status
@@ -13,6 +14,7 @@ from DashAI.back.api.api_v1.schemas.explainers_params import (
     ValidateDatasetParams,
     ValidDatasetsParams,
 )
+from DashAI.back.core.artifacts import GroupedArtifacts
 from DashAI.back.core.enums.status import ExplainerStatus
 from DashAI.back.dependencies.database.models import (
     Dataset,
@@ -24,6 +26,8 @@ from DashAI.back.dependencies.database.models import (
 
 if TYPE_CHECKING:
     from sqlalchemy.orm import sessionmaker
+
+    from DashAI.back.dependencies.registry.component_registry import ComponentRegistry
 
 logging.basicConfig(level=logging.DEBUG)
 log = logging.getLogger(__name__)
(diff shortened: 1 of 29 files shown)


cristian-tamblay (changes requested):
### Blocking: explanations created before v0.9.7 will 500 on plot load

`_attach_stories` walks the raw pickle and its normalized counterpart with
`zip(..., strict=True)`, which assumes the pickle is always a list. That only
holds for data written after `explainer_job.py` started normalizing before
pickling (`58c6262dc`, 2026-07-10). Tag v0.9.6 is from 2026-06-29, so any
explanation generated on 0.9.5/0.9.6 is still stored in its raw shape: a bare
plotly string or a bare dict.

Running the PR code against those shapes:

```
raw = bare plotly string   ->  ValueError: zip() argument 2 is shorter than argument 1
raw = bare dict            ->  ValueError: zip() argument 2 is shorter than argument 1
raw = bare dict (local)    ->  KeyError: 0
raw = list (current data)  ->  OK
```

When `raw` is a string, `zip` iterates its characters; when it is a dict, it
iterates its keys. The `KeyError: 0` comes from `_is_grouped_raw(raw[0])`.

Neither call site catches it (`explainers.py:491` and `:790`), so it surfaces
as a 500. These explanations render fine today because `normalize_artifacts`
handles the legacy shapes, so this is a regression for existing users.

Fix: mirror `normalize_artifacts`' own wrapping at the top of `_attach_stories`

```python
if isinstance(raw, (str, dict, Artifact, GroupedArtifacts)):
    raw = [raw]
```

and wrap the call site in a `try/except`, so the "never raises" contract in the
docstring actually holds and any future mismatch degrades to "no story" instead
of taking down the plot.

`test_explainer_story_attach.py` only covers the current normalized shape.
Worth adding the two legacy cases.

cristian-tamblay (approved):

PR #811, opened 2026-08-11
Document requirements and add AVX2 warning for AppImage

Irozuku:
## Summary
Legacy CPUs without AVX2 (pre 2013 Intel, pre Excavator AMD) get no usable signal from the desktop installers: the app starts, runs the database migration, then dies with a bare ``Illegal instruction (core dumped)``, because the bundled PyTorch binaries assume AVX2. Reported in #802 on an Intel i5-3210M (Ivy Bridge, 2012).

This adds the missing information in two places: a Requirements table at the top of the README covering OS, CPU and Python per install method, and a startup check in the AppImage entrypoint that names the cause and points at the source install before the crash happens.

---

## Type of Change

- [ ] Backend change
- [ ] Frontend change
- [ ] CI / Workflow change
- [ ] Bug fix
- [x] Documentation

---

## Changes (by file)
- `README.rst`: new Requirements section with a per install method table (operating system, CPU, Python). Documents the AVX2 requirement for the desktop installers, and that the PyPI / source install does not need it. Corrects the macOS entry to Apple Silicon only, since the Intel DMG job is `continue-on-error` and its artifact is never staged into the release (`publish.yml:187`, `publish.yml:300-342`). Adds a note in the installers section about the `Illegal instruction` crash.
- `appimage/entrypoint.sh`: AVX2 check before the `exec`, placed after the terminal relaunch so the message is visible in the log window. Prints the reason and the `uv sync --extra cpu` workaround.

---

## Testing
The check is advisory and gated on three conditions, so it should stay silent on normal hardware:

    $ grep -c avx2 /proc/cpuinfo     # non zero on any 2013+ CPU, no warning expected

To see the warning path on a machine that does have AVX2, force the branch by pointing the grep at an empty file, or run under a QEMU guest with `-cpu qemu64` (this is not recommended given qemu64 is a generic cpu that is in fact really slow, so setting up dashAI would take a lot of time and effort, pointing to an empty file is recommended given this case).

---

## Notes
- The check never exits: it only prints to stderr and continues. A missing CPUID bit can also come from a VM that masks it while AVX2 still works, and blocking a working install is worse than an occasional false warning. The message says so explicitly.
- Guards: only runs on `x86_64` (ARM has no `avx2` flag and would always warn), only when `/proc/cpuinfo` is readable, and matches `\bavx2\b` so substrings do not produce a false pass.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -16,6 +16,28 @@ AI models
 .. image:: ./images/dashai-logo.svg
    :alt: dashAI Logo
 
+Requirements
+============
+
+.. list-table::
+   :header-rows: 1
+   :widths: 20 40 40
+
+   * -
+     - Desktop installers
+     - PyPI / source install
+   * - **Operating system**
+     - Windows 10/11 x64, macOS 14+ on Apple Silicon (no Intel Mac build),
+       Linux x64 with glibc 2.35 or newer and FUSE 2
+     - Windows, macOS or Linux x64
+   * - **CPU**
+     - x86_64 **with AVX2** (Intel Haswell 2013+, AMD Excavator / Zen+), or
+       Apple Silicon
+     - x86_64 or Apple Silicon, AVX2 not required
+   * - **Python**
+     - Nothing to install: the installers bundle their own Python 3.12
+     - You install it yourself: Python 3.10 or greater (3.12 recommended)
+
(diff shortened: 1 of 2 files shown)


cristian-tamblay (approved):
Thank you for your report

PR #867, opened 2026-09-07
Declarative schema rules, and the search space as a declared type

cristian-tamblay:
## Summary

Schema fields could not depend on each other. A rule like "the partitions sum to at most 1" had no home, so it lived in a bespoke form, in prose nobody enforced, or nowhere. This branch gives that kind of statement a place to live: rules are declared as data next to the fields they constrain, published on the wire under `x-dashai-rules`, and evaluated by two interpreters — one in Python, one in JS — kept honest by a generated conformance fixture both consume.

The same idea then applies to the second thing a field could not declare: a hyperparameter's search space. That was carried by the *shape of its placeholder dict*, which is why HPO worked for ints and floats only, why 31 fields claimed to be optimizable and could not be, and why no component's schema accepted its own declared defaults.

Most of what this enables was already reachable and switched off. `enumNames` had been the key the renderer reads for dropdown labels since forever and nothing produced it, so every dropdown in the product showed `friedman_mse` in all five languages. `multipleOf` was described in prose instead of declared. Options scikit-learn rejects were being offered. A field emptied by the user sent `""` where a number was wanted.

**Nothing here is a rewrite.** Rule emission is additive, the four-key search envelope already in the database is unchanged, and no component or plugin is required to declare anything.

---

## Type of Change

- [x] Backend change
- [x] Frontend change
- [ ] CI / Workflow change
- [x] Bug fix
- [x] Documentation

---

## Changes (by file)

**The rule layer**

- `DashAI/back/core/schema_fields/rules.py` (new): the authoring API. `Check` and `Relevance`, and a closed 16-node expression AST (`F`, `Ctx`, `Sum`, `Approx`, `And`, `In`, `Len`, …) with operator overloads. Being a closed set of nodes rather than a grammar is what removes any parser-parity risk between the two runtimes. `validate_rules` refuses a rule that cannot fire, so a mistake fails at import instead of silently doing nothing.
- `DashAI/back/core/schema_fields/rules_eval.py` (new): the Python interpreter. Verdicts are three-valued — a `PENDING` sentinel whose `__bool__` raises, so a half-filled form cannot be read as a violation.
- `DashAI/front/src/utils/ruleEngine.js` (new): the mirror interpreter plus `withRules`, which folds the rules into the Yup schema so violations arrive through the channel formik already reads.
- `DashAI/back/core/schema_fields/base_schema.py`: merges parent and own rules, validates them at class definition, publishes them under `x-dashai-rules`, and runs them as an inherited validator.
- `tests/back/core/test_rules_conformance.py` (new): generates `ruleConformance.fixture.json` (1,241 cases) from the Python side; the JS suite consumes the same file, so the two interpreters cannot drift apart quietly.

**The search space**

- `DashAI/back/core/schema_fields/search_space.py` (new): `search_space(inner, fixed=, low=, high=, choices=)` makes the space the field's type. The placeholder is derived, so it cannot disagree with the field it describes, and `fixed`/`low`/`high`/every choice are checked against the field's own constraints when the class is defined.
- `DashAI/back/core/schema_fields/optimizer_{int,float}_field.py`: **deleted.** They were byte-identical to `int_field` and `float_field`, which is exactly why the placeholder's shape ended up carrying the signal.
- `DashAI/back/optimizers/optuna_optimizer.py`: `suggest_categorical` arm, so an enum or a boolean can be searched.
- `DashAI/back/optimizers/hyperopt_optimizer.py`: same, plus two bugs. Its space builder was an `if`/`elif` with no `else`, so a parameter of any other kind was left out of the space and *silently never optimized*. And `fmin` reports the **index** of a chosen option, which the code read through `float(raw_value)` — that would have set `loss = 1.0` on the final model. All three read sites now go through `space_eval`.
- `DashAI/back/models/model_factory.py`: reads a declared `x-dashai-search-dtype` instead of the JSON Schema `type`, which is absent whenever a field admits null and cannot say "categorical" at all.
- `DashAI/front/src/components/configurableObject/Inputs/SelectInputOptimize.jsx` (new): the third optimizer input, sibling to the two numeric ones. An interval makes no sense for something picked out of a set, so this offers the options instead.

**Forms that were available and switched off**

- `DashAI/back/core/schema_fields/enum_labels.py` (new): a hand-written table of 53 option-set vocabularies, 154 names in five languages, which now name 79 of the 217 enum fields in the tree. `enumNames` had been the key the renderer reads and was emitted **zero** times across 219 components, so every dropdown showed the raw Python value in every language. The 217 fields use only 111 distinct option sets, so the name belongs to the option set rather than to the field, which is what makes this a table rather than 126 edits. Sets whose meaning is not the same everywhere stay out on purpose: `("auto",)` means "pick the iteration count for me" in one field and "resample every class but the largest" in another.
- `DashAI/front/src/utils/schema.js`: `emptyValueFor` derives what an emptied field means from its own schema, which is the elegant home the `"" → None` patch never had; `applyMultipleOf`; bounds now read from the `anyOf` branch too.
- `DashAI/front/src/components/shared/FormSchemaRenderFields.jsx`: relevance effects, and `isOptimizable` now tested before `anyOf` — the union picker used to win, which is a second reason a nullable field could never show its optimize toggle.
- `DashAI/front/src/utils/columnEligibility.js` (new): one implementation of the column metadata contract, replacing three that disagreed.
- Leaf inputs and `SelectInput.jsx`: `disabled` support, and `""` coerced to null only when `""` is not itself an option (Plotly's `histnorm` offers it as a real value meaning raw counts, and coercing it made that option unselectable).

**Schemas migrated**

167 search spaces across 31 components (69 integer, 50 number, 48 categorical); relevance rules in 5 splitters sharing one rule object, `SimpleImputer`, `ExponentialSmoothing` and the covariance explorer; `multipleOf` in 6 diffusion schemas; option fixes in 4 sklearn models.

---

## Testing

Backend **2740 pass** (1 skip, 1 xfail); frontend **208 pass** (up from 187 — there were no form tests at all before, because `react-markdown` is ESM-only and CRA's jest does not transform `node_modules`, so importing any form component was a parse error).

Two things worth a reviewer's eye, both about failing loudly:

- `withRules` in `ruleEngine.js` uses Yup's `{is, then, otherwise}` form deliberately. The bare-function builder is called as `fn(...values, schema, options)`, so reading the schema off the end of the argument list picks up `options`, Yup throws "conditions must return a schema object", and **formik turns that into no errors at all**. There is a guard that fails loudly rather than silently disabling every validation.
- An unsupported search dtype is now a `TypeError`. `study.optimize` catches `ValueError` as an unfittable trial, so it used to surface as "every one of the N trials failed, narrow the ranges and try again" — the wrong advice for a declaration naming a kind of space that does not exist.

---

## Notes

**The audit of the 31 went both ways.** The fields declared `none_type(optimizer_int_field(...))` could not be optimized however they were declared. Read by name, 16 are `random_state` and 2 are `n_jobs`. A seed does not have a better value, it has a value that happens to score better on this split. So 13 were revived with a declared range and 18 demoted to ordinary fields — and 3 more of the same mistake turned up *live* (`RandomForestClassifier.random_state` was searchable over `0..10`, `verbose` over `0..100` on two models).

**Two declared ranges cannot produce a converged model, and I did not change them.** Both are on SVC: `max_iter` is searched over `-1..10` where the other twelve models offering it search 50..10000, and `-1` means unlimited, so the range mixes "no limit" with "ten iterations" as if they were points on one scale. `tol` is searched over `1.0..10.0` where eight siblings search `1e-6..0.1`, and `LogisticRegression.tol` starts at `0.0`. These pass every constraint, so no mechanism objects; changing a search range changes results, so they want a decision rather than a patch.

**Deliberately left as prose, with a test saying so.** The two `constant` converters depend on how many columns the user selected, which is session state and not a schema field: that needs the context channel. `ExponentialSmoothing.seasonal` cannot be searched because a `Relevance` rule reads it, and a rule over a search envelope can only ever be pending — `validate_rules` refuses such a rule, so trying stops the class from importing. That guard working is why the omission is a decision and not an oversight.

**Known and unfixed, found while measuring:** `TrainNode.jsx:476` compares partition sums with exact equality, so 60/30/10 is unsaveable; `validator.py:89` hardcodes the holdout splitter names; `utils/schema.js` gives an `anyOf` field `Yup.mixed().nullable()`, leaving 241 properties unvalidated; `applyMinMax` maps `exclusiveMinimum` to `.min()` at 12 sites; `IntegerInput.jsx:24` parses `"1e3"` as 1.

**Plugins:** the plugin guide documented `int_field`, `float_field`, `enum_field` and `schema_field`, never the two deleted factories, so a plugin following it is unaffected. One importing them directly breaks at import; a 6-line deprecated alias would give a grace period if that is wanted.

**Still open by design:** the optimizer's own UI, and `run.parameters` is not validated at read time, so the new runtime checks only bite where something calls `model_validate`.


DashAI/back/core/schema_fields/enum_labels.py
cristian-tamblay, line 1: @Irozuku  This is the enumName list we talked about

Irozuku (approved):

